PR #17, opened 2015-02-09
Removed EmptyCache field from DNSCache

### LINKED ISSUE #15: JDK 8 compatibility
Compiling against the JDK 8 libraries fails due to the return type of of javax.jmdns.impl.DnsCache._EmptyCache.keySet(). The return type was changed from Set<K> to ConcurrentHashMap.KeySetView<K,V>. 

I considered a few fixes, but couldn't come up with a solution that wouldn't either a) break builds against earlier JDK versions or b) introduce a breaking API change.

How strongly do you feel about API compatibility for the impl.* classes?

beowulfe:
Java 8 changed the signature of ConcurrentHashMap.keySet(), breaking compilation of javax.jmdns.impl.DNSCache. Considering that there are no consumers of this field within the project, and the class that contains it is in a impl package, removing it is the simplest option.
Fixes #15

Signed-off-by: Andy Lintner <dev@beowulfe.com> (github: beowulfe)

### DIFF
--- a/src/main/java/javax/jmdns/impl/DNSCache.java
+++ b/src/main/java/javax/jmdns/impl/DNSCache.java
@@ -9,8 +9,6 @@ import java.util.Collection;
 import java.util.Collections;
 import java.util.Iterator;
 import java.util.List;
-import java.util.Map;
-import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 
 import javax.jmdns.impl.constants.DNSRecordClass;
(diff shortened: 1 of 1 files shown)


beowulfe:
Removing it was the cleanest solution - it looks like it was probably leftover code from an earlier revision, before it was converted to use ConcurrentHashMap, but I didn't trace the history. As far as optimizations go, it's not a super useful one - there's only one DNSCache created for each instance of JmDNSImpl.

I did a cursory search of open source projects using this library and none are directly accessing this field. It feels safe to me, but if someone has a good use case for providing the static empty, I can revisit this with a more complex solution.

spearway:
I agree this is really just a left over from old habits…

Pierre

kaikreuzer:
Great, thanks!
So this was the only issue for Java8? Does all compile nicely now?

beowulfe:
Yes, compiles cleanly in Java8 now.

PR #23, opened 2015-07-31
fixes resource leaks from opened streams

kaikreuzer:


maggu2810:
What do you think, should we try to support and use try-with-resources?

kaikreuzer:
No, because if you look at http://stackoverflow.com/questions/tagged/jmdns, you will see that most people use jmDNS on Android - so we should stick to Java 1.6 language syntax!

maggu2810:
Ok, just to kept this note for android here:
* try-with-resources is supported if your minSdkVersion is set to 19 or higher
* The NDSManager (API level >= 16) should support mDNS service discovery

kaikreuzer:
ok @maggu2810 , you've got a point here - it is somehow difficult to understand how the Android and Java language syntaxes relate over the versions...
But I think we are on the safe side with the current code.

PR #24, opened 2015-07-31
updated and cleaned project resources and switched test dependencies …

kaikreuzer:
…to current versions.


.gitignore
```diff
@@ -1,2 +1,3 @@
 *.class
 target/*
+/target/
```
maggu2810: Would prefer using one line: **/target/

PR #26, opened 2015-08-04
removed command line docu, deactivated tests on a default maven build

kaikreuzer:


src/test/java/javax/jmdns/test/JmDNSTest.java
```diff
@@ -431,6 +432,7 @@ public boolean waitForReceive() {
     private final static int MPORT = 8053;
 
     @Test
+    @Ignore
```
janossch, line 435: @kaikreuzer Do you happen to remember why it was set to `@Ignored`?

kaikreuzer: No, sorry, I don't. Possibly because none of them were working...

PR #29, opened 2015-09-24
Allow customisation of JmDNS object creation

littlefreaky:
Added Factory class and Delegate interface to allow to customise JmDNS object creation (same as in JmmDNS and NetworkTopologyDiscovery). Without this, JmmDNS will always create a JmDNSImpl instance without a name. This will cause a call to InetAddress.getHostName() which does a reverse DNS lookup. This may be _very_ slow and therefore it takes a long time until the services are published to a new network interface. Providing a custom Delegate allows to create JmDNS objects with a name which will prevent this.

### DIFF
--- a/src/main/java/javax/jmdns/JmDNS.java
+++ b/src/main/java/javax/jmdns/JmDNS.java
@@ -9,6 +9,7 @@ import java.io.IOException;
 import java.net.InetAddress;
 import java.util.Collection;
 import java.util.Map;
+import java.util.concurrent.atomic.AtomicReference;
 
 import javax.jmdns.impl.JmDNSImpl;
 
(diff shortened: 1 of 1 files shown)


kaikreuzer:
> This will cause a call to InetAddress.getHostName() which does a reverse DNS lookup.

This is indeed an annoying feature. 

> Providing a custom Delegate allows to create JmDNS objects with a name which will prevent this.

Would you maybe want to add such an alternative implementation to this PR as well?

littlefreaky:
> Would you maybe want to add such an alternative implementation to this PR as well?

I'm not sure how such an alternative implementation should look like. My own implementation automatically generates a name if none is provided. But the name generation is specific to my use case, so this would not be a generic implementation that others could use.
It may be possible to add an interface whose implementations are responsible for generating the name. A generic Delegate implementation could then take such a name generator as constructor parameter and call back to it when a new name must be generated. 
Then again, since the JmDNS.Factory.ClassDelegate interface only contains one method it can be represented by a lambda. So implementing a generic class whose only function is to call the callback interface and then instantiate a JmDNSImpl object seems to be a bit verbose if you can just do

```java
JmDNS.Factory.setClassDelegate((addr, name) -> new JmDNSImpl(addr, generateName()));
```
Or if you could  reference a static method which does the job with

```java
JmDNS.Factory.setClassDelegate(MyClass::createJmDNS);

private static final JmDNS createJmDNS(InetAddress addr, String name)
{
  return new JmDNSImpl(addr, generateName());
}
```

What do you think?

kaikreuzer:
Looking a bit deeper into the code, I am just wondering, whether it is at all ok to have implementation classes besides JmDNSImpl. The code does not seem as if it cleanly uses the JmDNS interface, but it often does a cast to JmDNSImpl (see e.g. https://github.com/openhab/jmdns/blob/24c28e141abb68590cb8263cd6da95485eb4cb83/src/main/java/javax/jmdns/impl/ServiceEventImpl.java#L121).
Won't this lead to a ClassCastException with your change?

littlefreaky:
Yes, you are correct. In these cases the change could lead to a ClassCastException. The question is: Is it possible to remove those casts? It should never be necessary to cast down to the concrete implementation. If there are functions in JmDNSImpl which are needed but are not in the JmDNS interface, we may have to consider adding them to the interface.
I see that in the case you mentioned (ServiceEventImpl), the constructor takes an object of type JmDNSImpl but just passes it to the super constructor which takes a JmDNS object. So this would be an example of uneccessary narrowing of the type.

littlefreaky:
For the time being I could also adapt the JmDNS.Factory.ClassDelegate.newJmDNS method to return an object of type JmDNSImpl. This would force users of this new interface to create JmDNSImpl objects or subclass it.

kaikreuzer:
> Is it possible to remove those casts?

In theory, it should, in practice, I have no idea what will happen and where it might break.

> This would force users of this new interface to create JmDNSImpl objects or subclass it.

So why not simply build it into the JmDNSImpl itself? E.g. adding a new static method "setName" and use this name if the method had been called and otherwise call  InetAddress.getHostName()?

littlefreaky:
> So why not simply build it into the JmDNSImpl itself? E.g. adding a new static method 
> "setName" and use this name if the method had been called and otherwise call
> InetAddress.getHostName()?

I think a static setName(String) method would not be suitable because the name is used in the logging messages and therefore should be unique for every JmDNS instance IMO. It would be possible to add a setNameGenerator(JmDNSNameGenerator) method instead, but I would refrain from that idea because I don't like static/global state in Java classes. 
Also, the factory approach has the advantage that one could create a subclass and override existing methods of the JmDNSImpl class if the need arises.

maggu2810:
I would prefer to first try to remove all casts of the interface to the
impl class in a separate PR. This could reviewed and tested separately.
Using the interface should be enough in the most cases, perhaps there is
some desigm broken, too.

kaikreuzer:
Closing this PR, as it can lead to ClassCastExceptions as discussed.

PR #30, opened 2015-09-30
Support for Apple Home sharing

jkiddo:
I've added the dirty details for getting JmDNS to support the way Apple uses Zeroconf to discover shared libraries in iOS Remote


src/main/java/javax/jmdns/impl/DNSQuestion.java
```diff
@@ -271,16 +273,25 @@ boolean answeredBy(DNSEntry rec) {
...
                 answers.addAll(jmDNSImpl.getLocalHost().answers(this.getRecordClass(), DNSRecordClass.UNIQUE, DNSConstants.DNS_TTL));
                 answers.addAll(info.answers(this.getRecordClass(), DNSRecordClass.UNIQUE, DNSConstants.DNS_TTL, jmDNSImpl.getLocalHost()));
             }
+            final String constructedType = "_home-sharing._" + info.getProtocol() + "." + info.getDomain() + ".";
+            final String constructedName = "_" + info.getPropertyString("hG") + "._sub." + info.getTypeWithSubtype();
+            if (this.getType().equalsIgnoreCase(constructedType) && info.getPropertyString("hG") != null && this.getName().equalsIgnoreCase(constructedName)) {
+                final javax.jmdns.impl.DNSRecord.Pointer ptr = new DNSRecord.Pointer(constructedName, DNSRecordClass.CLASS_IN, DNSRecordClass.UNIQUE, DNSConstants.DNS_TTL, info.getName() + "." + constructedType);
+                final List<DNSRecord.Pointer> l = new LinkedList<DNSRecord.Pointer>();
+                l.add(ptr);
+                answers.addAll(l);
```
kaikreuzer: why not directly add ptr here?

pom.xml
```diff
@@ -1,7 +1,7 @@
 <project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
 	xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
 	<modelVersion>4.0.0</modelVersion>
-	<groupId>org.openhab.jmdns</groupId>
```
kaikreuzer: please leave the Maven groupId unchanged. We decided to change it to org.openhab in order to have full control for future release distributions, see https://github.com/openhab/jmdns/issues/6#issuecomment-143203737.

jkiddo: Sure ... my bad... that one should not have gone into the request. So in essence, does that mean that the original javax.jmdns project is deprecated?

src/main/java/javax/jmdns/impl/DNSQuestion.java
jkiddo: You're right ... I wrote it in a pretty late hour ... that should explain it ;)

pom.xml
kaikreuzer: > So in essence, does that mean that the original javax.jmdns project is deprecated?

Yes, there won't be any further releases done from Sourceforge (and thus the javax.jmdns Maven GroupId) anymore. All further evolution is done right here at Github.

kaikreuzer: @jkiddo Could you update your PR accordingly?

jkiddo:
All should be good to go now

kaikreuzer:
Thanks! One "final" question: Do you insist on adding all the "final" keywords in the code (which has nothing to do with the feature of this PR itself)? Usually, the Java compiler anyhow optimizes the bytecode, so that there is no need to have these keywords in the sources here - they rather make the code more difficult to read. So my preference would be to remove them again.

Could you in the end please also squash all commits into one and sign it off (see https://github.com/openhab/jmdns/blob/master/CONTRIBUTING.md#sign-your-work)?

jkiddo:
I've removed all the finals  introduced now an on-wards from there all commits will be signed of. I'm afraid that I'm to green in git to be able to squash the commits together. Could you make an exception as I've kept my branch updated with the master?

kaikreuzer:
> Could you make an exception as I've kept my branch updated with the master?

I'm afraid not as it would stain the commit history, but I can try to help you getting things done correctly.

Before this, just another question: Is there a reason why you haven't added the code in https://github.com/openhab/jmdns/blob/master/src/main/java/javax/jmdns/impl/ServiceInfoImpl.java#L1222? This would seem to be the more natural place at first sight.
Then again, there currently is no specific handling for any service type - your PR is introducing the very first case of such a special handling. Do you have any reference on the web about why this is necessary? Is Apple really not adhering to the mDNS specification here (which would be weird as it was them to create it in the first place)? Would be great to have some reference (possibly directly as a comment within the code) that makes it clear, why this code is necessary.

jkiddo:
I would have to check up on that...

I know that the PR I've made is highly vendor dependent . The reason the code is placed where it is placed is because it was what I currently could get working. I'm no expert on mDNS - I have focused on decompiling the Apple protocol and I would be willing to give it a shot for moving the code to the place where you suggest.

AFAIK, I'm the first one to have implemented this pseudo-Apple solution in my project, that you are able to use with the official Apple Remote App without having to exchange a PIN code - And the only way I initially could get it to work was to do the hacking in JmDNS.

kaikreuzer:
> And the only way I initially could get it to work was to do the hacking in JmDNS.

That's why I am asking for some background information, so that others can assess whether this change makes sense in general.
I am hesitant to accept "hacks" that are done for a specific project only and which might only be of value there and which might potentially have unwanted side effects on others.

jkiddo:
That is fully understandable. The conclusion is probably that this PR is premature in its current state. In terms of documentation, from what I can see in Wireshark and the behvaviour in the Apple Remote App is that I've added the code at the correct place.

kaikreuzer:
So shall we leave that PR unmerged for the moment and wait for others to come across it and comment? If there is clearly a problem that needs to be solved in general, we can then merge it any time.

jkiddo:
Thats fine by me

kaikreuzer:
Closing due to inactivity.

PR #32, opened 2015-10-28
Fixed handling of TXT records without '='

martenbr:
This commit fixes an issue with parsing TXT records without '=' in them.
The result of the issue is that the whole TXT record will be
considered invalid and the service will appear not have any properties
at all unless you use the raw byte array.

Signed-off-by: Mårten Borgström <martenborgstrom@gmail.com> (github: martenbr)


kaikreuzer:
Thanks!

PR #33, opened 2015-10-30
Prevent unnecessary network traffic during cache cleanup

talksinmath:
First of all, thanks for taking ownership of jmdns - it's really great to see it live again here.

We'd like to conritubte an enhancement that we recently implemented because we encountered the following problem:

In our environment there a a lot of service instances (for the same service type,  >30) and a lot of listeners for these service (> 20). Under certain circumstances we saw MDNS traffic on our network exceeding 1 MB/s (caused by our services).

Close investigation revealed the following:
Since some of service instances were sometimes not shutdown correctly (no "Goodbye"-Packets), these remained in the cache of all the listeners. During cache cleanup (every 10s) each listener would try to renew this service once the TTL of this record is less than 50% of its original value (60 min).

This leads to the following behaviour:
After 30 min, every listener would query every 10 s for the service type of the stale record and never receive an answer from this record (and therefore never resetting the TTL). But every available service instance would answer every query from every listener nonetheless....
So we have about 20 listeners asking every 10 s with more than 30 service instances answering until the TTL of the stale/dead record has elapsed. Additionally, we've seen multiple network request for a single service type in each refresh cycle, which also adds network load because every request is answered by all available services.

This seemed to be the cause of the above mentioned network flooding.

According to the RFC in [section 5.2](https://tools.ietf.org/html/rfc6762#section-5.2) the following behaviour should be implemented:
>  "To perform this cache maintenance, a Multicast DNS querier should
>   plan to retransmit its query after at least 50% of the record
>   lifetime has elapsed.  This document recommends the following
>   specific strategy.
>
>   The querier should plan to issue a query at 80% of the record
>   lifetime, and then if no answer is received, at 85%, 90%, and 95%.
>   If an answer is received, then the remaining TTL is reset to the
>   value given in the answer, and this process repeats for as long as
>   the Multicast DNS querier has an ongoing interest in the record.  If
>   no answer is received after four queries, the record is deleted when
>   it reaches 100% of its lifetime.  A Multicast DNS querier MUST NOT
>   perform this cache maintenance for records for which it has no local
>   clients with an active interest.  If the expiry of a particular
>   record from the cache would result in no net effect to any client
>   software running on the querier device, and no visible effect to the
>   human user, then there is no reason for the Multicast DNS querier to
>   waste network capacity checking whether the record remains valid.
>
>   To avoid the case where multiple Multicast DNS queriers on a network
>   all issue their queries simultaneously, a random variation of 2% of
>   the record TTL should be added, so that queries are scheduled to be
>   performed at 80-82%, 85-87%, 90-92%, and then 95-97% of the TTL."
   
This pull request contains our implementation of this behaviour.
We are looking forward to your review.
Thanks again! :)

### DIFF
--- a/src/main/java/javax/jmdns/impl/DNSRecord.java
+++ b/src/main/java/javax/jmdns/impl/DNSRecord.java
@@ -13,6 +13,7 @@ import java.net.InetAddress;
 import java.net.UnknownHostException;
 import java.util.HashMap;
 import java.util.Map;
+import java.util.Random;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 
@@ -33,6 +34,8 @@ public abstract class DNSRecord extends DNSEntry {
     private static Logger logger = Logger.getLogger(DNSRecord.class.getName());
     private int           _ttl;
     private long          _created;
+    private int           _isStaleAndShouldBeRefreshedPercentage;
+    private final int     _randomStaleRefreshOffset;
 
     /**
      * This source is mainly for debugging purposes, should be the address that sent this record.
@@ -46,6 +49,8 @@ public abstract class DNSRecord extends DNSEntry {
         super(name, type, recordClass, unique);
         this._ttl = ttl;
         this._created = System.currentTimeMillis();
+        _randomStaleRefreshOffset = new Random().nextInt(3);
+        _isStaleAndShouldBeRefreshedPercentage = DNSConstants.STALE_REFRESH_STARTING_PERCENTAGE + _randomStaleRefreshOffset;
     }
 
     /*
@@ -150,11 +155,33 @@ public abstract class DNSRecord extends DNSEntry {
     }
 
     /**
+     * Check if the record is stale and whether the record should be refreshed over the network.
+     *
+     * @param now
+     *            update date
+     * @return <code>true</code> is the record is stale and should be refreshed, <code>false</code> otherwise.
+     */
+    public boolean isStaleAndShouldBeRefreshed(long now) {
+        return getExpirationTime(_isStaleAndShouldBeRefreshedPercentage) <= now;
+    }
+
+    /*
+    * Increment the percentage that determines whether a record needs to be refreshed.
+     */
+    void incrementRefreshPercentage() {
+        _isStaleAndShouldBeRefreshedPercentage += DNSConstants.STALE_REFRESH_INCREMENT;
+        if (_isStaleAndShouldBeRefreshedPercentage > 100) {
+            _isStaleAndShouldBeRefreshedPercentage = 100;
+        }
+    }
+
+    /**
      * Reset the TTL of a record. This avoids having to update the entire record in the cache.
      */
     void resetTTL(DNSRecord other) {
         _created = other._created;
         _ttl = other._ttl;
+        _isStaleAndShouldBeRefreshedPercentage = DNSConstants.STALE_REFRESH_STARTING_PERCENTAGE + _randomStaleRefreshOffset;
     }
 
     /**
--- a/pom.xml
+++ b/pom.xml
@@ -14,8 +14,18 @@ The project was originally started in December 2002 by Arthur van Hoff at Strang
 		<name>JmDNS</name>
 		<url>http://jmdns.org/</url>
 	</organization>
+	<developers>
+		<developer>
+			<name>Kai Kreuzer</name>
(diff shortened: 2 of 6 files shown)


src/main/java/javax/jmdns/impl/DNSRecord.java
```diff
@@ -150,11 +155,33 @@ public boolean isStale(long now) {
...
+        }
+    }
+
+    /**
      * Reset the TTL of a record. This avoids having to update the entire record in the cache.
      */
     void resetTTL(DNSRecord other) {
         _created = other._created;
         _ttl = other._ttl;
+        _isStaleAndShouldBeRefreshedPercentage = DNSConstants.STALE_REFRESH_STARTING_PERCENTAGE + _randomStaleRefreshOffset;
```
Flowdalic: Shouldn't this be ` + **other**._randomStaleRefreshedPercentage`?

talksinmath: Thanks for the input, I can see your point. In principle this should not matter since it is a random value that is only used to spread the network load over time.

I would actually prefer to change the signature of `resetTTL(DNSRecord)` to `resetTTL(created, ttl)` but I did not want to perform changes that were not necessary for our enhancement.

Do you agree?

Flowdalic: > In principle this should not matter since it is a random value that is only used to spread the network load over time.

Right, but then you could have just omitted that line. As it stands now it's a NOP.

> I would actually prefer to change the signature of resetTTL(DNSRecord) to resetTTL(created, ttl)

That decision is up to the jmdns@openhab maintainers. I would probably deprecate `resetTTL(DNSRecord)` and add `resetTTL(long, int)`.

talksinmath: I think we have a misunderstanding about what I meant when I said it does not matter which random value gets used ('this' or 'other').

It is absolutely crucial that is gets reset to the 'STALE_REFRESH_STARTING_PERCENTAGE'. It is just the random offset that does not need to be copied, since the DNSRecord instance already has its own. 
It is not a NOP, omitting it would break the resetting of the stale check interval.

Flowdalic: > I think we have a misunderstanding

You are right. I didn't notice that _isStaleAndShouldBeRefreshedPercentage is modified later on.

kaikreuzer:
> First of all, thanks for taking ownership of jmdns - it's really great to see it live again here.

Well, since the original authors have left, I think it is really important that the project has a home, where others can still maintain it - so I am trying to help as good as I can. Not being in the mDNS specs in depth myself, I can only do cursory reviews, so I rely on others to peer-review PRs (therefore thanks to @Flowdalic for looking at it as well!).

From what I see, the PR lgtm. I would only ask you to
- add a unit test if possible
- sign off your commit (see https://github.com/openhab/jmdns/blob/master/CONTRIBUTING.md#sign-your-work)

Thanks!
Kai

talksinmath:
I tried to add the sign-off, hope it worked.

kaikreuzer:
> hope it worked

Only partially - we now have 3 instead of 1 commit, so you would need to squash them.
Will you also add a test?

talksinmath:
I can add a unit-test that tests the simple behavior of a DNSRecord instance but not a full-fledged integration test for the whole time-out behavior. Would that suffice?

As for the squashing - could you give me a hint how to go about that? I'm fairly new to git and starting to feel lost...

kaikreuzer:
> Would that suffice?

Yes!

> could you give me a hint how to go about that?

This should hopefully help you: https://github.com/ginatrapani/todo.txt-android/wiki/Squash-All-Commits-Related-to-a-Single-Issue-into-a-Single-Commit

talksinmath:
Thanks for your help, but I am not getting anywhere with this squash. No matter what I do, I'm not getting the results I expect (trying Github Desktop, Sourcetree, cli). I'm feeling I'm making a mess of this :(

Since I will need to add a unit test, it is probably best for me to start this pull request from scratch.
Truly sorry about this, but git brought me to my knees....

kaikreuzer:
sure, feel free to create a new PR. Btw, you should always create PRs from a branch, never from your master.

talksinmath:
Will do. Thanks :)

PR #37, opened 2015-11-05
Fix building issues

taiyangc:
Two things are addressed here:

1) "ant jar" works again by removing duplicate classes and old tools. It's probably no longer used, but FWIW.

2) JmDNS toString() does not crash anymore after building with Java 8, something similar to #17.

Signed-off-by: Eric Chen <taiyang.chen@gmail.com> (github: taiyangc)

### DIFF
--- a/build.xml
+++ b/build.xml
@@ -58,9 +58,9 @@
 		</javac>
 	</target>
 
-	<target name="jar" depends="build,tools" description="Jar the files">
+	<target name="jar" depends="build" description="Jar the files">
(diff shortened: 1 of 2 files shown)


kaikreuzer:
Thanks for the fixes. May I ask you to sign off your commits before I merge (see https://github.com/openhab/jmdns/blob/master/CONTRIBUTING.md#sign-your-work)?

taiyangc:
Commits are now signed off.

kaikreuzer:
Thank you!

PR #38, opened 2015-11-09
Prevent unnecessary network traffic during cache cleanup, with test

talksinmath:
Signed-off-by: Carsten <carsten@meyermuc.de>

We'd like to conritubte an enhancement that we recently implemented because we encountered the following problem:

In our environment there a a lot of service instances (for the same service type, >30) and a lot of listeners for these service (> 20). Under certain circumstances we saw MDNS traffic on our network exceeding 1 MB/s (caused by our services).

Close investigation revealed the following:
Since some of service instances were sometimes not shutdown correctly (no "Goodbye"-Packets), these remained in the cache of all the listeners. During cache cleanup (every 10s) each listener would try to renew this service once the TTL of this record is less than 50% of its original value (60 min).

This leads to the following behaviour:
After 30 min, every listener would query every 10 s for the service type of the stale record and never receive an answer from this record (and therefore never resetting the TTL). But every available service instance would answer every query from every listener nonetheless....
So we have about 20 listeners asking every 10 s with more than 30 service instances answering until the TTL of the stale/dead record has elapsed. Additionally, we've seen multiple network request for a single service type in each refresh cycle, which also adds network load because every request is answered by all available services.

This seemed to be the cause of the above mentioned network flooding.

According to the RFC in section 5.2 the following behaviour should be implemented:

>To perform this cache maintenance, a Multicast DNS querier should
>plan to retransmit its query after at least 50% of the record
>lifetime has elapsed. This document recommends the following
>specific strategy.
>
>The querier should plan to issue a query at 80% of the record
>lifetime, and then if no answer is received, at 85%, 90%, and 95%.
>If an answer is received, then the remaining TTL is reset to the
>value given in the answer, and this process repeats for as long as
>the Multicast DNS querier has an ongoing interest in the record. If
>no answer is received after four queries, the record is deleted when
>it reaches 100% of its lifetime. A Multicast DNS querier MUST NOT
>perform this cache maintenance for records for which it has no local
>clients with an active interest. If the expiry of a particular
>record from the cache would result in no net effect to any client
>software running on the querier device, and no visible effect to the
>human user, then there is no reason for the Multicast DNS querier to
>waste network capacity checking whether the record remains valid.
>
>To avoid the case where multiple Multicast DNS queriers on a network
>all issue their queries simultaneously, a random variation of 2% of
>the record TTL should be added, so that queries are scheduled to be
>performed at 80-82%, 85-87%, 90-92%, and then 95-97% of the TTL."

This pull request contains our implementation of this behaviour.
We are looking forward to your review.
Thanks again! :)

### DIFF
--- a/src/main/java/javax/jmdns/impl/DNSRecord.java
+++ b/src/main/java/javax/jmdns/impl/DNSRecord.java
@@ -13,6 +13,7 @@ import java.net.InetAddress;
 import java.net.UnknownHostException;
 import java.util.HashMap;
 import java.util.Map;
+import java.util.Random;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 
@@ -33,6 +34,8 @@ public abstract class DNSRecord extends DNSEntry {
     private static Logger logger = Logger.getLogger(DNSRecord.class.getName());
     private int           _ttl;
     private long          _created;
+    private int           _isStaleAndShouldBeRefreshedPercentage;
+    private final int     _randomStaleRefreshOffset;
 
     /**
      * This source is mainly for debugging purposes, should be the address that sent this record.
@@ -46,6 +49,8 @@ public abstract class DNSRecord extends DNSEntry {
         super(name, type, recordClass, unique);
         this._ttl = ttl;
         this._created = System.currentTimeMillis();
+        _randomStaleRefreshOffset = new Random().nextInt(3);
+        _isStaleAndShouldBeRefreshedPercentage = DNSConstants.STALE_REFRESH_STARTING_PERCENTAGE + _randomStaleRefreshOffset;
     }
 
     /*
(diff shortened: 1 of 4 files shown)


kaikreuzer:
Looks all good now - thanks!

talksinmath:
Great! Right now we are using our own build of jmdns- do you have any timetable for a next release? I would love to be able to ditch ours and switch to the official releases. Thanks :)

kaikreuzer:
3.4.2 isn't that long ago yet, I would wait for a few more fixes to get into the code base.
So keep on fixing stuff and the sooner I will be willing to do a new release ;-)

talksinmath:
Sounds good. Been waiting for a release since 2012 anyways ;)

kaikreuzer:
So I hope you had seen https://github.com/openhab/jmdns/releases/tag/jmdns-3.4.2 in August already!

talksinmath:
I did see it too late. Actually found you here when we were investigating the problem that lead to this pull request, and since we need the fix we will use our own version and wait for the next release to switch to it - but that's not really a problem. I'm just glad the project has a new home!

PR #43, opened 2016-01-22
Set checks if it contains itself. Fixed to check for name parameter.

cpore:
Minor change, should be evident that wrong variable is used.


kaikreuzer:
Looks indeed evident. Thanks!

PR #72, opened 2016-05-17
Host Name Cannot Be User-Defined

danieljanev:
It seems that there is a bug in the javax.jmdns.impl.HostInfo class and its newHostInfo(InetAddress address, JmDNSImpl dns, String jmdnsName) method. In my project it is mandatory to export a gateway as mDNS device with host name using the following pattern:

Miele-.local.

So we are using the following code for the creation of the JmDNS instance for exporting the Gateway:

jmDNS = JmDNS.create(InetAddress.getLocalHost(), hostName);

where the hostName is created according the pattern described above. Using the library the Gateway was always exported with the Host name of the box but not the one which is passed as parameter. I found something which looks like a bug in the newHostInfo method. Here is the original source:

```
public static HostInfo newHostInfo(InetAddress address, JmDNSImpl dns, String jmdnsName) {
    HostInfo localhost = null;
    String aName = (jmdnsName != null ? jmdnsName : "");
    InetAddress addr = address;
    try {
        if (addr == null) {
            String ip = System.getProperty("net.mdns.interface");
            if (ip != null) {
                addr = InetAddress.getByName(ip);
            } else {
                addr = InetAddress.getLocalHost();
                if (addr.isLoopbackAddress()) {
                    // Find local address that isn't a loopback address
                    InetAddress[] addresses = NetworkTopologyDiscovery.Factory.getInstance().getInetAddresses();
                    if (addresses.length > 0) {
                        addr = addresses[0];
                    }
                }
            }
            if (addr.isLoopbackAddress()) {
                logger.warn("Could not find any address beside the loopback.");
            }
        }
        if (aName.length() == 0) {
            aName = addr.getHostName();
        }
        if (aName.contains("in-addr.arpa") || (aName.equals(addr.getHostAddress()))) {
            aName = ((jmdnsName != null) && (jmdnsName.length() > 0) ? jmdnsName : addr.getHostAddress());
        }
        logger.warn("Could not intialize the host network interface on " + address + "because of an error: " + e.getMessage(), e);
        // This is only used for running unit test on Debian / Ubuntu
        addr = loopbackAddress();
        aName = ((jmdnsName != null) && (jmdnsName.length() > 0) ? jmdnsName : "computer");
    }
    // A host name with "." is illegal. so strip off everything and append .local.
    // We also need to be carefull that the .local may already be there
    int index = aName.indexOf(".local");
    if (index > 0) {
    }
    aName = aName.replaceAll("[:%\\.]", "-");
    aName += ".local.";
    localhost = new HostInfo(addr, aName, dns);
    return localhost;
}
```

and here is the fix which we've done:

```
public static HostInfo newHostInfo(InetAddress address, JmDNSImpl dns, String jmdnsName) {
    HostInfo localhost = null;
    String aName = (jmdnsName != null ? jmdnsName : "");
    InetAddress addr = address;
    try {
        if (addr == null) {
            String ip = System.getProperty("net.mdns.interface");
            if (ip != null) {
                addr = InetAddress.getByName(ip);
            } else {
                addr = InetAddress.getLocalHost();
                if (addr.isLoopbackAddress()) {
                    // Find local address that isn't a loopback address
                    InetAddress[] addresses = NetworkTopologyDiscovery.Factory.getInstance().getInetAddresses();
                    if (addresses.length > 0) {
                        addr = addresses[0];
                    }
                }
            }
            if (addr.isLoopbackAddress()) {
                logger.warn("Could not find any address beside the loopback.");
            }
        }
        if (aName.length() == 0) {
            aName = addr.getHostName();
        }
        if (aName.contains("in-addr.arpa") || (aName.equals(addr.getHostName()))) {
            aName = ((jmdnsName != null) && (jmdnsName.length() > 0) ? jmdnsName : addr.getHostAddress());
        }
        logger.warn("Could not intialize the host network interface on " + address + "because of an error: " + e.getMessage(), e);
        // This is only used for running unit test on Debian / Ubuntu
        addr = loopbackAddress();
        aName = ((jmdnsName != null) && (jmdnsName.length() > 0) ? jmdnsName : "computer");
    }
    // A host name with "." is illegal. so strip off everything and append .local.
    // We also need to be carefull that the .local may already be there
    int index = aName.indexOf(".local");
    if (index > 0) {
    }
    aName = aName.replaceAll("[:%\\.]", "-");
    aName += ".local.";
    localhost = new HostInfo(addr, aName, dns);
    return localhost;
}
```

Using this fix now my gateway is always exported using the pattern I want to use. I'd like to ask you to check whether this is really a bug or I'm not using the correct methods.

### DIFF
--- a/src/main/java/javax/jmdns/impl/HostInfo.java
+++ b/src/main/java/javax/jmdns/impl/HostInfo.java
@@ -89,7 +89,7 @@ public class HostInfo implements DNSStatefulObject {
             if (aName.length() == 0) {
                 aName = addr.getHostName();
             }
-            if (aName.contains("in-addr.arpa") || (aName.equals(addr.getHostAddress()))) {
+            if (aName.contains("in-addr.arpa") || (aName.equals(addr.getHostName()))) {
                 aName = ((jmdnsName != null) && (jmdnsName.length() > 0) ? jmdnsName : addr.getHostAddress());
             }
         } catch (final IOException e) {


kaikreuzer:
Hi @danieljanev,

Sorry for coming back so late on this.
Before merging, I tried to dive into the code to understand the change - which I actually fail to do.

I have written this test:
```
    @Test
    public void testNewHostInfo() throws IOException {
        String mDNSName = "Miele-.local.";
        JmDNS jmdns = JmDNS.create(InetAddress.getLocalHost(), mDNSName);
        assertEquals(jmdns.getHostName(), mDNSName);        
    }
```
which is green, even without your change. So what exactly is the problem you are trying to address? Would be great if you could add an according unit test to your PR, which is red on the current code and becomes green with your changes.

danieljanev:
Hi @kaikreuzer 

The main problem is that the mDNS device is exported in the network using the host name of the machine but not the specified one. E.g. on QIVICON box the name is QIVICON.local. but not Miele-1234567890.local.

kaikreuzer:
How to capture this in a test? Even if I do
```
        String mDNSName = "Miele-.local.";
        JmDNS jmdns = JmDNS.create(InetAddress.getByName("127.0.0.1"), mDNSName);
        assertEquals(jmdns.getHostName(), mDNSName);
        
        jmdns = JmDNS.create(InetAddress.getByName("127.0.0.1"), null);
        assertEquals(jmdns.getHostName(), "localhost.local.");
```
everything is green. And it shows that the "name" parameter is clearly taken into account, even if a hostname for the ip address can be resolved.

danieljanev:
I'm not sure we can check it with a test. We found this problem using the JmDNS browser :(

kaikreuzer:
Well, if you claim that the change in HostInfo fixes a bug, you should be able to specify which input to this method results in which wrong output.

kaikreuzer:
@danieljanev Any success in your further analysis? If not, I would suggest to close this PR for the moment.

danieljanev:
@kaikreuzer I'll try create a test till the end of the week. Too busy with some internal product work these days ...

kaikreuzer:
I plan to do a new JmDNS patch release on Thursday - so if you want this in, better do it mid than end of the week ;-) Otherwise, we will schedule it for the next one.

kaikreuzer:
@danieljanev What is the status on this? Are you still convinced that there is a bug in the code? If so, let us address this as I plan to do a new patch release soon!

kaikreuzer:
@danieljanev Would be great if you could briefly comment!

LionisIAm:
Agree with issue starter.
You can see that Service name equals to the target name. "SlavkaAndroid._airplay._tcp.local:" to the "target SlavkaAndroid.local" what is not good.

```
SlavkaAndroid._airplay._tcp.local: type SRV, class IN, cache flush, priority 0, weight 0, port 7000, target SlavkaAndroid.local
    Service: SlavkaAndroid
    Protocol: _airplay
    Name: _tcp.local
    Type: SRV (Server Selection) (33)
    .000 0000 0000 0001 = Class: IN (0x0001)
    1... .... .... .... = Cache flush: True
    Time to live: 3599
    Data length: 22
    Priority: 0
    Weight: 0
    Port: 7000
    Target: SlavkaAndroid.local

```

 We have to get something like:
```
SlavkaDesktop._airplay._tcp.local: type SRV, class IN, cache flush, priority 0, weight 0, port 7000, target DESKTOP-1UVTC9H.local
            Service: SlavkaDesktop
            Protocol: _airplay
            Name: _tcp.local
            Type: SRV (Server Selection) (33)
            .000 0000 0000 0001 = Class: IN (0x0001)
            1... .... .... .... = Cache flush: True
            Time to live: 120
            Data length: 24
            Priority: 0
            Weight: 0
            Port: 7000
            Target: DESKTOP-1UVTC9H.local
```
I`m sure you can see the difference. Big problem for me as for now.
Thanks

sobimibos:
We made an investigation on this issue. Tests and code analysis show, that the existing code is complete correct in regard of this issue.
Additionally, the proposed change does not change anything in this situation, because the effected if-clause is not executed anyway if using a custom name, not without the change and not with the change.

kaikreuzer:
Thanks @sobimibos, this was also the result of my analysis, see above.

@danieljanev & @LionisIAm It would be hugely helpful if you could come up with a test case that clearly shows your issue and which can prove that the suggested change fixes it. I think you agree that by looking at the change, it simply does not seem to make any kind of sense.

kaikreuzer:
As there is no feedback since summer 2016, I am closing this PR now.
Once you have figured out a test case and can explain the suggested change, feel free to re-open it.

PR #83, opened 2016-07-28
Use default TTL instead of DNSConstants

FMesmer:
By setting DNSStateTask.setDefaultTTL(newDefaultTTLValue) a new default TTL can be set.
But in the code the value of the constant is still used.
It would be helpful if at least in the DNSQuestion class in the method addAnswersForServiceInfo the "DNSConstants.DNS_TTL" value is replaced by "DNSStateTask.defaultTTL()".
With these changes the answer of an DNS question uses the new default TTL value.

As mentioned above their exists similar methods still using the constant value. But I am not sure if all occurrences should be replaced.

Regards
Frank

### DIFF
--- a/src/test/java/javax/jmdns/test/JmDNSTest.java
+++ b/src/test/java/javax/jmdns/test/JmDNSTest.java
@@ -10,6 +10,7 @@ import java.net.InetAddress;
 import java.net.MulticastSocket;
 import java.net.NetworkInterface;
 import java.net.UnknownHostException;
+import java.util.Collection;
 import java.util.Enumeration;
 import java.util.HashMap;
 import java.util.Map;
@@ -19,7 +20,13 @@ import javax.jmdns.ServiceEvent;
 import javax.jmdns.ServiceInfo;
 import javax.jmdns.ServiceListener;
 import javax.jmdns.ServiceTypeListener;
+import javax.jmdns.impl.DNSCache;
+import javax.jmdns.impl.DNSEntry;
+import javax.jmdns.impl.DNSRecord;
+import javax.jmdns.impl.JmDNSImpl;
 import javax.jmdns.impl.constants.DNSConstants;
+import javax.jmdns.impl.constants.DNSRecordType;
+import javax.jmdns.impl.tasks.state.DNSStateTask;
 
 import org.easymock.Capture;
 import org.easymock.EasyMock;
@@ -517,6 +524,30 @@ public class JmDNSTest {
             Thread.sleep(1500);
             services = registry.list(service.getType());
             assertTrue("We should not see the service we just unregistered: ", services == null || services.length == 0);
+        } finally {
+            if (registry != null) registry.close();
+        }
+    }
+
+    @Test
+    public void testRegisterServiceWithDifferentDefaultTLL() throws IOException, InterruptedException {
+        System.out.println("Unit Test: testRegisterServiceWithDifferentDefaultTLL()");
+        JmDNS registry = null;
+        int defaultTTL = 30;
+        try {
+            DNSStateTask.setDefaultTTL(defaultTTL);
+            registry = JmDNS.create();
+            registry.registerService(service);
+
+            JmDNSImpl registryImpl = (JmDNSImpl) registry;
+            DNSCache cache = registryImpl.getCache();
+            Collection<DNSEntry> entries = cache.allValues();
+            for (DNSEntry entry : entries) {
+                DNSRecord record = (DNSRecord) entry;
+                if (record.getRecordType() == DNSRecordType.TYPE_PTR || record.getRecordType() == DNSRecordType.TYPE_SRV || record.getRecordType() == DNSRecordType.TYPE_TXT) {
+                    assertEquals(record.getRecordType().toString(), defaultTTL, record.getTTL());
+                }
+            }
         } finally {
             if (registry != null) registry.close();
         }
--- a/src/test/java/javax/jmdns/test/JmDNSTest.java
+++ b/src/test/java/javax/jmdns/test/JmDNSTest.java
@@ -530,16 +530,17 @@ public class JmDNSTest {
     }
 
     @Test
-    public void testRegisterServiceWithDifferentDefaultTLL() throws IOException, InterruptedException {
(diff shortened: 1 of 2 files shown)


src/test/java/javax/jmdns/test/JmDNSTest.java
```diff
@@ -521,4 +528,30 @@ public void testListMyServiceWithoutLowerCase() throws IOException, InterruptedE
             if (registry != null) registry.close();
         }
     }
+
+    @Test
+    public void testRegisterServiceWithDifferentDefaultTLL() throws IOException, InterruptedException {
```
kaikreuzer: TTL, not TLL

src/test/java/javax/jmdns/test/JmDNSTest.java
```diff
@@ -521,4 +528,30 @@ public void testListMyServiceWithoutLowerCase() throws IOException, InterruptedE
...
+        System.out.println("Unit Test: testRegisterServiceWithDifferentDefaultTLL()");
```
kaikreuzer: TTL, not TLL

src/test/java/javax/jmdns/test/JmDNSTest.java
```diff
@@ -521,4 +528,30 @@ public void testListMyServiceWithoutLowerCase() throws IOException, InterruptedE
...
+        JmDNS registry = null;
+        int defaultTTL = 30;
+        try {
+            DNSStateTask.setDefaultTTL(defaultTTL);
+            registry = JmDNS.create();
+            registry.registerService(service);
+            // Needed - otherwise DNSQuestion.addAnswersForServiceInfo is not called
+            ServiceInfo[] services = registry.list(service.getType());
```
kaikreuzer: Right, but you can remove the `ServiceInfo[] services = ` to avoid "unused" warnings in the IDE.

src/test/java/javax/jmdns/test/JmDNSTest.java
```diff
@@ -521,4 +528,30 @@ public void testListMyServiceWithoutLowerCase() throws IOException, InterruptedE
...
+            Collection<DNSEntry> entries = cache.allValues();
+            for (DNSEntry entry : entries) {
+                DNSRecord record = (DNSRecord) entry;
+                if (record.getRecordType() == DNSRecordType.TYPE_PTR || record.getRecordType() == DNSRecordType.TYPE_SRV || record.getRecordType() == DNSRecordType.TYPE_TXT) {
+                    assertTrue(record.getRecordType().toString(), Math.abs(defaultTTL - record.getTTL()) <= 1);
+                }
+            }
+        } finally {
+            if (registry != null) registry.close();
+        }
```
kaikreuzer: You should probably also do a 
```
            DNSStateTask.setDefaultTTL(DNSConstants.DNS_TTL);
```
at the end.

kaikreuzer:
Hi Frank,

Thanks for the PR. Trying to read through the code, I am actually not quite sure which problem you exactly want to address here. The method `DNSStateTask.setDefaultTTL()` is not meant to be used by productive code, hence it will always return `DNSConstants.DNS_TTL` and can only be used within some unit tests. But since you do not change any existing unit test, this does not seem to be your goal. So what exactly is this supposed to do?

FMesmer:
Hy Kai,

we need another default TTL instead of 60 minutes. So the only way I found was the DNSStateTask.setDefaultTTL() to do this. So I want this call in our productive code.
By testing I found that there are several methods which add or change the cache with its DNS entries. Sometimes the defaultTTL was used often just the constant. 
Maybe its clearer if you tried the test without the changes in the DNSQuestion class. And comment/uncomment the line ServiceInfo[] services = registry.list(service.getType()); to see the differences in the cache. With the DNSQuestion changes the defaultTTL value is used for the entries.

Regards Frank

kaikreuzer:
> we need another default TTL instead of 60 minutes

As far as I am aware, the default of 60 minutes is defined in the specification, therefore I would say it must not be changed.

If I understand it correct, DNSQuestion are classes for INCOMING messages, right? So if an external service does not specify a custom TTL, it is absolutely correct to use the default TTL according to the spec. If you don't like this, you might rather have to talk to the IETF to change the spec instead...?

FMesmer:
Hy Kai,
I am not really in the code yet. But as I tested the code, DNSQuestion answer the question (also maybe) about its own services. So here the TTL is insert to give the time to live for its own service. 
In the specification given in the project I found a minimum of 2 minutes and an optimum at 75 minutes.
My request is about if someone wants to change the TTL, the changed TTL is given to other devices. In the moment changing the TTL by setDefaultTTL() other devices get still the 60 minutes.

In the next days I have a closer look at the code. Because we have also problems with double announcements of the services. These false announcements can not be removed.

Greetings
Frank

kaikreuzer:
> So here the TTL is insert to give the time to live for its own service. 

Ok, but then the right thing to do is to introduce a mechanism which allows to define a custom TTL for this service (instead of using the default TTL as a fallback). You might actually also want to be able to set different TTLs for different services that you provide.
So using the setDefaultTTL() method (where its JavaDocs says that it must not be used productively) is definitely not a valid solution, I hope you agree. So I'd suggest to close this PR for now - looking forward to a follow up, once you have had a closer look!

FMesmer:
Ok.
In the moment I read the draft-cheshire-dnsext-multicastdns.txt

10.  Resource Record TTL Values and Cache Coherency

   As a general rule, the recommended TTL value for Multicast DNS
   resource records with a host name as the resource record's name (e.g.
   A, AAAA, HINFO, etc.) or a host name contained within the resource
   record's rdata (e.g. SRV, reverse mapping PTR record, etc.) SHOULD be
   120 seconds.

   The recommended TTL value for other Multicast DNS resource records is
   75 minutes.
------
I will compare this with other mDNS implementation and jmdns. 

"10.3.  Cache Flush on Topology change" is about flush cache records. Here I don't see interfaces in jmdns for this. I will check this further and will make a next pull request for this.

Regards
Frank

kaikreuzer:
@spearway, In case you have any input on that matter, please feel invited to share your wisdom!

PR #86, opened 2016-08-02
Only allow a listener to be added once.

### LINKED ISSUE #85: addServiceListerner in JmDNSImpl does not check correctly that the same listener is already added
If you call addServiceListerner several times with the same object, the object will be added several times and later the object will be notified several times.

kaikreuzer:
fixes #85

### DIFF
--- a/src/main/java/javax/jmdns/impl/JmDNSImpl.java
+++ b/src/main/java/javax/jmdns/impl/JmDNSImpl.java
@@ -52,7 +52,7 @@ import javax.jmdns.impl.util.NamedThreadFactory;
 /**
  * mDNS implementation in Java.
  *
- * @author Arthur van Hoff, Rick Blair, Jeff Sonstein, Werner Randelshofer, Pierre Frisch, Scott Lewis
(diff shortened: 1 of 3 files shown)


stefaneicher:
Hi Kai, I could run the code and saw that the test fails in case i change the line "synchronized (list) {"
But I do not have a deeper knowledge about the code and therefore can not (or not yet) give you a more valuable feedback.

FYI: Intelij gave a warning in the old code.
bildschirmfoto 2016-08-02 um 21 03 09

In fact Intellij reports a lot of warning and erros. Almost all seem to to valuable. I think going over those warning could bring a lot.

Let me know if I can help with something else.

kaikreuzer:
Thanks, that's good enough for me to feel good about merging. I just wanted to obey the 4-eyes-principle, just in case I made any stupid mistakes.

> a lot of warning and erros. Almost all seem to to valuable.

If you find issues that are clearly bugs, please create PRs for them!

Note that I am not at all into the codebase myself, all that I am trying to do here is to keep the project alive, since there are quite some people using it. But I would definitely not dare any bigger refactoring (although I am often tempted to do so when I look at the code ;-)).

PR #88, opened 2016-08-03
Type jar is default

stefaneicher:
type jar is default and must not be set

Signed-off-by: Stefan Eicher <stefan.eicher@gmail.com> (github: stefaneicher)


kaikreuzer:
You have accidentially included two other commits in this PR. Could you rebase your branch?

PR #90, opened 2016-08-04
type jar is default and must not be set

stefaneicher:
type jar is default and must not be set

Signed-off-by: Stefan Eicher <stefan.eicher@gmail.com> (github: stefaneicher)


stefaneicher:
please delete the other PR. I need to get into the GIT stuff ...

PR #98, opened 2016-10-07
Feature for DNS Constants and bugfix for DNSTaskStarter

dschroe:
- DNS TTL is now configurable via System Properties for faster removal detection, default value is still one  
  hour
- Timer for DNS cache is fixed and marked again as non daemon thread

### DIFF
--- a/src/main/java/javax/jmdns/impl/DNSTaskStarter.java
+++ b/src/main/java/javax/jmdns/impl/DNSTaskStarter.java
@@ -275,7 +275,7 @@ public interface DNSTaskStarter {
             super();
             _jmDNSImpl = jmDNSImpl;
             _timer = new StarterTimer("JmDNS(" + _jmDNSImpl.getName() + ").Timer", true);
-            _stateTimer = new StarterTimer("JmDNS(" + _jmDNSImpl.getName() + ").State.Timer", true);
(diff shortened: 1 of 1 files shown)


src/main/java/javax/jmdns/impl/constants/DNSConstants.java
```diff
@@ -15,9 +15,9 @@
     // changed to final class - jeffs
     public static final String MDNS_GROUP                     = "224.0.0.251";
     public static final String MDNS_GROUP_IPV6                = "FF02::FB";
-    public static final int    MDNS_PORT                      = Integer.parseInt(System.getProperty("net.mdns.port", "5353"));
+    public static final int    MDNS_PORT                      = Integer.getInteger("net.mdns.port", 5353);
     public static final int    DNS_PORT                       = 53;
-    public static final int    DNS_TTL                        = 60 * 60;                                                      // default one hour TTL
+    public static final int    DNS_TTL                        = Integer.getInteger("net.dns.ttl", 60 * 60);                   // default one hour TTL
     // public static final int DNS_TTL = 120 * 60; // two hour TTL (draft-cheshire-dnsext-multicastdns.txt ch 13)
```
kaikreuzer: Thanks for this PR, looks good to me (and sorry for the late response!).

Just one question: This comment correctly states that the TTL should be 2 hours according to the mDNS spec. Do you know why it is nonetheless set to 1 hour as a default in the code?

PR #101, opened 2016-11-03
Fix for computer with multi-interfaces

rmath:
Using HostName with JmmDNS + rejecting all queries / responses coming
from an other interface

### DIFF
--- a/src/main/java/javax/jmdns/impl/JmmDNSImpl.java
+++ b/src/main/java/javax/jmdns/impl/JmmDNSImpl.java
@@ -5,6 +5,7 @@ package javax.jmdns.impl;
 
 import java.io.IOException;
 import java.net.InetAddress;
+import java.net.UnknownHostException;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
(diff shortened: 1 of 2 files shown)


src/main/java/javax/jmdns/impl/JmmDNSImpl.java
```diff
@@ -85,11 +87,20 @@
...
+    private String                                              _defaultHostName = null;
+
     /**
      *
      */
     public JmmDNSImpl() {
         super();
+
+        try {
+            _defaultHostName = InetAddress.getLocalHost().getHostName();
```
spearway: This actually makes a DNS query which may not be advisable in this class.

rmath: Thanks for your response. I will try another way to set the same name for all JmDNS instances of the same computer.

kaikreuzer: @rmath Do you still plan to do an update here?

kaikreuzer:
Closing due to inactivity.

PR #113, opened 2017-05-10
Bugfix/DNSRecord.Text off by one

ViToni:
While changing `toString()` methods of `DNSRecord` subclasses to avoid the creation of additional `String`s a small issue was found which showed a (in most cases) unprintable character befor the content of `DNSRecord.Text`.

This was due to `DNSRecord.Text` having a `byte` array which contains the length of the contained data at `index 0`, so this first `byte` shouldn't be used for `String`  creation.

### DIFF
--- a/src/main/java/javax/jmdns/impl/DNSIncoming.java
+++ b/src/main/java/javax/jmdns/impl/DNSIncoming.java
@@ -350,11 +350,31 @@ public final class DNSIncoming extends DNSMessage {
                 rec = new DNSRecord.Service(domain, recordClass, unique, ttl, priority, weight, port, target);
                 break;
             case TYPE_HINFO:
-                StringBuilder buf = new StringBuilder();
-                buf.append(_messageInputStream.readUTF(len));
-                int index = buf.indexOf(" ");
-                String cpu = (index > 0 ? buf.substring(0, index) : buf.toString()).trim();
-                String os = (index > 0 ? buf.substring(index + 1) : "").trim();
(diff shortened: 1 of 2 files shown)


kaikreuzer:
> This was due to DNSRecord.Text having a byte array which contains the length of the contained data at index 0, so this first byte shouldn't be used for String creation.

Where does this length byte come from? Is this defined in the specification? Is this an implementation detail (and thus a bug that it is not considered here)? Is it a remote device sending records that do not conform to the specification?
I'd just like to understand what exactly we are fixing here.

Would there be a chance to also add a unit test for this?

ViToni:
The raw data coming from the socket contains the length of the data as the first `byte`.
This raw data is stored without modification as a `byte` array into `DNSRecord.Text`.
This is perfectly fine.

The bug is only in the way the `toString()` method was implemented here, beacuse it uses this first `byte` as part of the `String`.

In other parts of the code this first `byte` is correctly handled, e.g.
https://github.com/jmdns/jmdns/blob/b1a4ccc01acc78893a1ca9f5d40d53d8d359960f/src/main/java/javax/jmdns/impl/DNSIncoming.java#L166
https://github.com/jmdns/jmdns/blob/b1a4ccc01acc78893a1ca9f5d40d53d8d359960f/src/main/java/javax/jmdns/impl/DNSIncoming.java#L118

ViToni:
Added one more related fix, because CPU/OS for HINFO weren't correctly  parsed out of the byte stream.

Issue #110, opened 2017-03-16
concurrentModificationException in registerServiceType() of JmDNSimpl.java


goingeast:
I saw this crash, why don't you synchronize the previous subtypes.contains()?

if ((subtypes != null) && (**!subtypes.contains(subtype)**)) {
                synchronized (subtypes) {
                    if (!subtypes.contains(subtype)) {
                        typeAdded = true;
                        subtypes.add(subtype);
                        final ServiceTypeListenerStatus[] list = _typeListeners.toArray(new ServiceTypeListenerStatus[_typeListeners.size()]);
                        final ServiceEvent event = new ServiceEventImpl(this, "_" + subtype + "._sub." + name, "", null);
                        for (final ServiceTypeListenerStatus status : list) {
                            _executor.submit(new Runnable() {
                                /** {@inheritDoc} */
                                @Override
                                public void run() {
                                    status.subTypeForServiceTypeAdded(event);
                                }
                            });
                        }
                    }
                }
            }

PR #117, opened 2017-05-16
Changes to avoid ConcurrentModificationExceptions.

ViToni:
Access to collections which might be modified concurrently has been synchronized.

Signed-off-by: Victor Toni <victor.toni@gmail.com>


ViToni:
This PR might solve #110.

Issue #120, opened 2017-05-20
Jmdns will crash when in Android AP mode.


psionic12:
Caused by: android.system.ErrnoException: setsockopt failed: ENODEV (No such device)
      at libcore.io.Posix.setsockoptGroupReq(Native Method)
      at libcore.io.ForwardingOs.setsockoptGroupReq(ForwardingOs.java:144)
      at libcore.io.IoBridge.setSocketOptionErrno(IoBridge.java:429)
      at libcore.io.IoBridge.setSocketOption(IoBridge.java:356)
      at java.net.PlainDatagramSocketImpl.setOption(PlainDatagramSocketImpl.java:189) 
      at java.net.PlainDatagramSocketImpl.join(PlainDatagramSocketImpl.java:128) 
      at java.net.MulticastSocket.joinGroup(MulticastSocket.java:149) 
      at javax.jmdns.impl.JmDNSImpl.openMulticastSocket(JmDNSImpl.java:473) 
      at javax.jmdns.impl.JmDNSImpl.<init>(JmDNSImpl.java:421) 
      at javax.jmdns.JmDNS.create(JmDNS.java:78) 
      at com.mobvoi.tichome.hsf.message.b$2.run(LocalNetworkChannel.java:101) 
      at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1112) 
      at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:587) 
      at java.lang.Thread.run(Thread.java:818)

ViToni:
As the surrounding code is not made available it's difficult to guess the cause for this exception.

Having to wild guess I'm curious if this branch changes anything for you:
https://github.com/ViToni/jmdns/tree/bugfix/crash_in_Android_AP_mode

postrcase:
Hmm, it doesn't seem to work. I try to discover a service when in AP mode. Built-in Android NSD works though (but has also a serious bug, that's why I try JmDNS).
I pass InetAddress.getLocalHost() to JmDNS.create. As a result, I get this exception:

W/DNSResolverTask: ServiceResolver(localhost.local.).run() exception 
                                                                     java.net.SocketException: sendto failed: ENETUNREACH (Network is unreachable)
                                                                         at libcore.io.IoBridge.maybeThrowAfterSendto(IoBridge.java:586)
                                                                         at libcore.io.IoBridge.sendto(IoBridge.java:555)
[...]

Issue #119, opened 2017-05-19
binds wildcard interface


andyli:
Similar to #69, when binding to a wildcard interface, e.g. 
```java
JmDNS jmdns = JmDNS.create(new InetSocketAddress(0).getAddress());
// ...
jmdns.registerService(serviceInfo);
```
clients will receive the `0.0.0.0` ip address but not the actual ip address.

Tested with JmDNS 3.5.1.

ViToni:
@andyli Could you please check with 3.5.3?

andyli:
I discontinued the project that used jmdns, and I don't think I can dig out the source code.
Feel free to close this if you think it has been fixed.

kaikreuzer:
Let's close it then.

PR #127, opened 2017-06-25
Use joinGroup() with specific network interface if available.

ViToni:
Using a specific network interface avoids the multicast group being joined on all interfaces.
This change is also needed to avoid crashing on Android when AP mode is active.


ViToni:
This PR is should solve:

https://github.com/jmdns/jmdns/issues/119
https://github.com/jmdns/jmdns/issues/120

PR #141, opened 2017-11-18
Some bad services can trigger null text value

Tolriq:
As per title.


kaikreuzer:
> I have another one to potentially make

Then you should be quick, I am just doing the 3.5.4 release!

> not sure what solution you'd prefer

Not sure either. Let's discuss on your branch.

PR #142, opened 2017-12-05
Bug fix jmDNS unable to cache-flush when the cache-flush bit is set

garytse:
Summary: Implementation did not follow closely to the RFC-6762 section 10.2 description and is fixed.

Description: we have tested that the jmdns is unable to cope with IP address change of one of our D-LINK camera 2330L.  This is because the jmdns library had a minor mistake in the cache-flush implementation.  It has be fixed to follow as described as RFC-6762 section 10.2

The fix is currently under testing.   The PR is created for early comments.  Will post again when tested on more devices.

Signed-off-by: Gary Tse <gary.tse@telekom.de>

### DIFF
--- a/src/main/java/javax/jmdns/impl/JmDNSImpl.java
+++ b/src/main/java/javax/jmdns/impl/JmDNSImpl.java
@@ -1346,14 +1346,22 @@ public class JmDNSImpl extends JmDNS implements DNSStatefulObject, DNSTaskStarte
             final boolean unique = newRecord.isUnique();
             final DNSRecord cachedRecord = (DNSRecord) this.getCache().getDNSEntry(newRecord);
             logger.debug("{} handle response cached record: {}", this.getName(), cachedRecord);
+            // RFC 6762, section 10.2 Announcements to Flush Outdated Cache Entries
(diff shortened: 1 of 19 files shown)


src/main/java/javax/jmdns/impl/JmDNSImpl.java
```diff
@@ -1346,14 +1346,21 @@ void handleRecord(DNSRecord record, long now) {
...
+            //     1. same name
+            //     2. same record type
+            //     3. same record class
+            //     4. record is older than 1 second.
             if (unique) {
                 for (DNSEntry entry : this.getCache().getDNSEntryList(newRecord.getKey())) {
                     if (    newRecord.getRecordType().equals(entry.getRecordType()) &&
                             newRecord.getRecordClass().equals(entry.getRecordClass()) &&
-                            (entry != cachedRecord)
+                            cachedRecord.getCreated() < (now - DNSConstants.FLUSH_RECORD_OLDER_THAN_1_SECOND*1000)
```
ViToni: I would prefer a method here, like eg. `boolean isOlderThanOneSecond(DNSEntry)` to make it more readable and expressive.

kaikreuzer:
Why did you close this PR?

PR #147, opened 2018-01-10
Bugfix Issue#95 stale ip address still reported + Fix warnings for listener events

garytse:
1. Issue#95 Fix cache-flush to follow mdns specification (RFC 6762, section 10.2 Announcements to Flush Outdated Cache Entries)

2. Fix warnings for listener events 
After fixing the cache-flush, ListenerStatus frequently fires new warnings "Service Resolved called for an unresolved event: {}".  This is caused by concurrent modification of the ServiceInfo object.


ViToni:
LGTM. Nice change!

Issue #145, opened 2017-12-24
JmDNS does not work on some multihomed machines


dimgit:
Hi,
I am using JmDNS 3.5.3, on Windows 7. In my setup I have 5 network interfaces - one physical and four virtual (three from VMs and one from a VPN).
In one of the VMs I have a service that I am discovering from the host.
I am not able to discover that service with the above setup. All the multicasts are sent to the VPN interface. I am not sure if the VPN software is to blame or not (it is Check Point Endpoint Security), though I have no network issues with any other software on that machine.
After looking through the code, specifically JmDNSImpl.java, I noticed that the multicast socket interface is not set, and instead it relies on the join() to receive messages only from the JmDNSImpl object interface.
I believe that is not enough. In my case it looks like the multicast may not be sent via the JmDNSImpl interface but instead via whatever the OS considers a default interface.
Debugging the code, I saw that the VPN interface receives the multicasts sent by the JmDNSImpl objects. That does not sound right.

If I add this line, before line 464 in JmDNSImpl.java, so the code looks like that:
`
_socket = new MulticastSocket(DNSConstants.MDNS_PORT);
_socket.setNetworkInterface(hostInfo.getInterface());
`

everything works Ok and I am able to discover the service on my VM.

Do you see any issues with setting the socket interface, as it used to be before?

Regards,

Dimitar

dhufnagel:
Hi,
I can confirm this issue. I also set the networkInterface on the socket and it works. I'll try to add a Pull request as soon as I can. But I first want to evaluate if there is also a problem on the announcement side of a service.

Regards,
Dominik

ViToni:
Could you please retest with master? I guess it should be solved now since #150 has been merged.

dimgit:
Hi,
I tested with the current master. The issue is fixed.
Thank you,
D.

ViToni:
@dimgit As it is resolved could you please close the issue.

dimgit:
Closing as the problem has been fixed.

PR #150, opened 2018-01-17
Added network interface to MulticastSocket (see #145)

dhufnagel:
Signed-off-by: Dominik Hufnagel d.hufi@gmx.de (github: dhufnagel)


ViToni:
LGTM.

src/main/java/javax/jmdns/impl/JmDNSImpl.java
```diff
@@ -468,6 +468,7 @@ private void openMulticastSocket(HostInfo hostInfo) throws IOException {
             logger.trace("Trying to joinGroup({}, {})", multicastAddr, hostInfo.getInterface());
 
             // this joinGroup() might be less surprisingly so this is the default
+            _socket.setNetworkInterface(hostInfo.getInterface());
```
kaikreuzer: Should this be better above the comment and also above the logging statement?

ViToni: Good point. Above the logging statement would be better.

Issue #164, opened 2018-08-29
Service is resolving without a hostname


jcgoforth:
In 3.5.4, on the service discovery side the service can become resolved without a hostname in the ServiceInfo (getServer() returns "").

The underlying issue is that the service is becoming resolved without ever receiving a SRV record. Additionally, even if it does receive the SRV record the hostname of the service ends up being dropped before the service becomes resolved.

jcgoforth:
I have a fix for this, but I am completely new to git and I am working on getting everything setup correctly for the pull request.

stef02:
can you please tell me about the fix?

senft:
Are there any updates on this? I am currently experiencing the same problem.

jcgoforth:
I dropped the ball on getting the pull request completed. I keep meaning to get back to it, I will try to get back to it this soon. In the mean time, this issue is fixed on my branch (https://github.com/jcgoforth/jmdns). I have been using a build of that code for the last 7 months without issue.

senft:
Thanks @jcgoforth, I will have a look at your fork.

janossch:
@kaikreuzer Imho this issue can be closed, since there is a closed PR specifically to this issue https://github.com/jmdns/jmdns/pull/165

@jcgoforth If you edit your main comment in your PR (https://github.com/jmdns/jmdns/pull/165), and you add the following line `Fix #164` as the first line it will link this PR with the ticket.

PR #165, opened 2018-08-29
Fix bug causing getServer() to always be empty (#164)

jcgoforth:
Fix a couple of bugs that caused ServiceInfo.getServer() to
always return an empty string when processing a resolved Service.
- ServiceInfoImpl.clone() didn't include _server
- ServiceInfoImpl.hasData() failed to properly check for _server being set

### DIFF
--- a/src/main/java/javax/jmdns/ServiceInfo.java
+++ b/src/main/java/javax/jmdns/ServiceInfo.java
@@ -436,10 +436,18 @@ public abstract class ServiceInfo implements Cloneable {
 
     /**
      * Get the name of the server.
+     * <b>Note:</b> This will be an empty string if no value has been set
      * 
      * @return server name
      */
     public abstract String getServer();
+
+    /**
+     * Returns true if the service info has a server
+     *
+     * @return <code>true</code> if the service info has a server specified, <code>false</code> otherwise.
+     */
+    public abstract boolean hasServer();
 
     /**
      * Returns the host IP address string in textual presentation.<br/>
--- a/src/main/java/javax/jmdns/impl/ServiceInfoImpl.java
+++ b/src/main/java/javax/jmdns/impl/ServiceInfoImpl.java
@@ -452,6 +452,14 @@ public class ServiceInfoImpl extends ServiceInfo implements DNSListener, DNSStat
     }
 
     /**
+     * @see javax.jmdns.ServiceInfo#hasServer()
+     */
+    @Override
+    public boolean hasServer() {
+        return (this._server != null);
+    }
+
+    /**
      * @param server
      *            the server to set
      */
@@ -894,7 +902,7 @@ public class ServiceInfoImpl extends ServiceInfo implements DNSListener, DNSStat
      * @param dnsCache
      * @param now
      * @param record to get data from
-     * @return
+     * @return <code>true</code> if service was updated, <code>false</code> otherwise
      */
     private boolean handleUpdateRecord(final DNSCache dnsCache, final long now, final DNSRecord record) {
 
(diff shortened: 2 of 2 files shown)


src/main/java/javax/jmdns/ServiceInfo.java
```diff
@@ -436,11 +436,19 @@ public static ServiceInfo create(final Map<Fields, String> qualifiedNameMap, fin
 
     /**
      * Get the name of the server.
+     * <b>Note:</b> This will be an empty string if no value has been set
```
kaikreuzer: The `_server` field is initialised in the constructor with the `text` parameter, so there should not be a such situation as "no value has been set` - and if there is: Maybe the right place to address it would be the constructor?

src/main/java/javax/jmdns/impl/ServiceInfoImpl.java
```diff
@@ -451,6 +451,14 @@ public String getServer() {
         return (_server != null ? _server : "");
```
kaikreuzer: Isn't this rather the bug here? It looks weird to return an empty string if no server is set... Did you manage to find out, why this is done this way?

src/main/java/javax/jmdns/impl/ServiceInfoImpl.java
```diff
@@ -1165,6 +1173,7 @@ public String getNiceTextString() {
     @Override
     public ServiceInfoImpl clone() {
         ServiceInfoImpl serviceInfo = new ServiceInfoImpl(this.getQualifiedNameMap(), _port, _weight, _priority, _persistent, _text);
+        serviceInfo.setServer(_server);
```
kaikreuzer: All other fields (besides the IP addresses) are set through the constructor and the constructor also already sets the `_server` field - so it looks a bit weird here to set this through a different mechanism. Wdyt?

src/main/java/javax/jmdns/ServiceInfo.java
jcgoforth: I may need to take another look at this. It would seem that the value of the `text` is itself an empty string or not valid input into the `ByteWrangler.encodeText()` method. I know that in my tests by the time the service was resolved the value of `_server == null`

src/main/java/javax/jmdns/impl/ServiceInfoImpl.java
jcgoforth: Perhaps it is, but given that it has been that was for 9 years I was hesitant to change it without fully understanding why it is there. I know that builds newer than that line of code have not had the particular bug I was trying to correct.

src/main/java/javax/jmdns/impl/ServiceInfoImpl.java
jcgoforth: I had considered adding it to the constructor but there is a little bit of weirdness here that drove me to do it this way. While true that the constructor initializes `_server` using the value of `text`, the value can then be updated when processsing the SRV record. If that has happened prior to calling this `clone()` we end up with a clone that isn't quite the same thing. I also can't just put the value of `_server` as the value in the `text` parameter since that value is encoded in the constructor and we could end up with a double encoded string.

IvanBoychevMarinov:
Hi! 
Although I can not say I have a huge knowledge about this topic I have few observations about it. First of all the copy constructor of the ServiceInfoImpl is not setting the _server. This constructor is used in the clone method of the "ServiceEventImpl".  The other constructor where the _server is not set at all is the "ServiceInfoImpl(Map<Fields, String> qualifiedNameMap, int port, int weight, int priority, boolean persistent, byte text[])" 

I'm not sure if the last one causes any issues but the copy constructor results in services with empty or null servers.

kaikreuzer (approved):
I must admit that I still do not fully grasp how the bug can happen, but on a second look the code changes at least do not seem to do any harm and if they prevent the bug and make sure that the cloned instances are correct, let's go for it.

Sorry to having that PR open for more than 2 years now - finally it'll end up in a release now!

Issue #144, opened 2017-12-20
Registering Service is rather slow


schulz-m:
We are currently evaluating different ZeroConfiguration Libraries and functionality-wise `jmdns` works rather nicely. But a big caveat is that the Service registering takes around 7 - 8 seconds, whereas with other libraries it seems to be well under a second. The culprit seems to be in the `waitForAnnounced` calls inside the [`registerService`](https://github.com/jmdns/jmdns/blob/master/src/main/java/javax/jmdns/impl/JmDNSImpl.java#L997) methods. Now I do understand that for a truly "save registering", you have to check if the service was announced in between, but why does it take so much longer than listening to existing Services with `ServiceListener`?

Any hints and explanations would be appreciated, especially in comparison to other mDNS/DNS-SD libraries out there :-)

We are using `jmdns` version `3.5.2` installed with maven.

mustafasezgin:
Im seeing this too on Android. Registering a service takes a fair while. Did you end up finding a solution to this @schulz-m ?

schulz-m:
@mustafasezgin Unfortunately we still haven't had time to look into this, will definitely keep it updated once we could!

cschneider:
I experimented a bit with this. 

It seems the main wait (~5s) happens when we do:
this.waitForAnnounced(DNSConstants.SERVICE_INFO_TIMEOUT);
Here it waits until the jmsDNS instance itself is announced.  I suspect that we somehow miss the first announce on start and then must wait until DNS announces itself again.

The other wait is after we register our service (~2S):
info.waitForAnnounced(DNSConstants.SERVICE_INFO_TIMEOUT);
Not sure why this takes so long.

I found two ways to make jmdns faster. 
The first is to remove the two waits in JmDNSImpl.registerService. The disadvantage is of course that we are not sure the service is announced but for me that would not be a problem. I am not sure though if this affects other parts of the system. 

The second way I found is to remove only 
this.waitForAnnounced(DNSConstants.SERVICE_INFO_TIMEOUT);
and put it into create.  This also speeds up the process a lot.

So my question is: Can we safely remove the waits? What purpose do they serve?

ViToni:
Both timeouts mentioned seem to be way too high. [RFC - Section 8.1](https://tools.ietf.org/html/rfc6762#section-8.1) describes the probing step to be complete (when no conflicts arise) after about 750ms.

kaikreuzer:
[...]

PR #172, opened 2018-10-12
Bug 170 fix the missing services resolved event sending

### LINKED ISSUE #170: Services that were removed are reported as resolved.
Sometimes a service that was removed gets reported as resolved at some later point. I think I have a fix for this, but I need you guys to take a look. Basically in the method "updateRecord" in the JmDNSImpl when a serviceInfo is about to be removed it is not removed from the listeners list which is basically a list of ServiceInfo instances. I believe that afterwards this ServiceInfo object is not removed at all and somtimes this results in sending Resolved events for this service info.

IvanBoychevMarinov:
As I described in the issue itself I'm getting resolved events for services already removed. I think this is because the listeners collection is not cleared but I need you guys to verify and confirm this solution is not breaking anything else.

Fixes #170


IvanBoychevMarinov:
Sorry, about the formatting. It was done automatically and I thought that it is not bad thing, so I decided to leave it this way. I will revert these right away.

About the solution... It really seems like it is doing the right thing but I was afraid I might be missing some other aspect that might got broken by this fix. 

I will remove the formatting and try to explain the fix afterwards .

IvanBoychevMarinov:
Ok, I think I was able to revert the formatting. Now there are few util classes used in the test, the test itslef and the fix. 

It seems that the javax.jmdns.impl.tasks.resolver.ServiceInfoResolver adds a "listener" to the JmDNSImpl. This listener is actually an instance of the "ServiceInfoImpl". This listener might get "removed" in the "cancel" method, but this won't happen if the ServceInfo is "persistent" which, according to the JavaDoc means that

>  _Returns <code>true</code> if ServiceListener.resolveService will be called whenever new new information is received._

 which seems ok until the ServiceInfo is there. I really think this "listener" aka the ServiceInfoImpl has to be removed when the ServiceInfo is removed. I have traced the chain of starting a service resolver and it seems a service resolver will be started if the service is added once again and it makes me feel that this is the right thing to do in this case.

Oh and the issue itself is caused by obsolete "listener" aka ServiceInfoImpl instance that invokes dns.handleServiceResolved in its updateRecord method.

IvanBoychevMarinov:
@kaikreuzer, can you please review the changes. The biggest change is in the JmdnsImpl file. The rest are just two helpful test classes and a test method in the JmdnsImplTest.

IvanBoychevMarinov:
@ViToni I saw you were the last one that modified the updateRecord method in the JmDNSImpl. Can you please take a look. 
To describe the issue in short, since the listeners were never stopped, they are not removed and thus they are constantly notified for service resolved even if those services are removed. You can see my previous comments for further info. The other changes are just unit tests demonstrating the behvaiour.

src/main/java/javax/jmdns/impl/JmDNSImpl.java
```diff
@@ -1018,12 +1018,15 @@ public void registerService(ServiceInfo infoAbstract) throws IOException {
         info.addAddress(_localHost.getInet4Address());
         info.addAddress(_localHost.getInet6Address());
 
+        this.waitForAnnounced(DNSConstants.SERVICE_INFO_TIMEOUT);
```
kaikreuzer, line 1021: How is this related to the bug?
Isn't this dangerous here as it blocks the method from returning?

src/main/java/javax/jmdns/impl/JmDNSImpl.java
```diff
@@ -1018,12 +1018,15 @@ public void registerService(ServiceInfo infoAbstract) throws IOException {
...
+
         this.makeServiceNameUnique(info);
         while (_services.putIfAbsent(info.getKey(), info) != null) {
             this.makeServiceNameUnique(info);
         }
 
         this.startProber();
+        info.waitForAnnounced(DNSConstants.SERVICE_INFO_TIMEOUT);
```
kaikreuzer, line 1029: Why a second time here?

src/test/java/javax/jmdns/impl/EventContainer.java
```diff
@@ -0,0 +1,54 @@
...
+import java.util.LinkedHashMap;
+import java.util.Map;
+
+import javax.jmdns.ServiceEvent;
+
+public class EventContainer {
+
+    private final Map<String, ServiceEvent> events = new LinkedHashMap<String, ServiceEvent>();
+
+    private boolean                         eventReceived;
```
kaikreuzer, line 12:
```suggestion
    private boolean eventReceived;
```

src/test/java/javax/jmdns/impl/JmDNSTest.java
```diff
@@ -557,4 +558,45 @@ public void testListMyServiceWithoutLowerCase() throws IOException, InterruptedE
             if (registry != null) registry.close();
         }
     }
+
+    @Test
+    public void shouldNotNotifyServiceListenersForServiceResolveAfterServiceRemoved() throws UnknownHostException, IOException, InterruptedException {
+        String firstType = "_test._type1.local.";
+        String secondType = "_test._type2.local.";
+
+        JmmDNS jmmdns = JmmDNS.Factory.getInstance();
```
kaikreuzer, line 567: We are here in the JmDNSTest class, not the JmmDNSTest class - I'd hence assume you should also create JmDNS instances here, right?

kaikreuzer (changes requested):
Hey @IvanBoychevMarinov,

Very sorry for having neglected that project for such a long time and your PR with it.
I'm just going through the open PRs right now and I hope you are still around and interested in following up on that one! Let me know, what you think wrt my comments below. Thanks!

src/main/java/javax/jmdns/impl/JmDNSImpl.java
IvanBoychevMarinov: Sorry, I will have to take a look to verify why is this shown as it is part of my commit. It is not. I presume it was there before my changes and I just moved it.

src/main/java/javax/jmdns/impl/JmDNSImpl.java
IvanBoychevMarinov: Sorry, I will have to take a look to verify why is this shown as it is part of my commit. It is not. I presume it was there before my changes and I just moved it.

src/test/java/javax/jmdns/impl/JmDNSTest.java
IvanBoychevMarinov: Sorry, for two years, I have to recall a lot of things. There a lot of things I do not remember. I will try to provide answers soon.

src/main/java/javax/jmdns/impl/JmDNSImpl.java
IvanBoychevMarinov: I did a quick check and this strings appears in two commits
* 9689c79aa270572a1f48878a52a93ccbe4aa3d9a -  Added timeout on waitForAnnounced
* fee0d4c27cdc159a1b4a0fa5883ddf562a696d00 -  Remove waits. Fixes #144 (#174)

I'm not sure why those waits got back in my commit, but I will remove it. This is the only non intended difference between my commit and fee0d4c27cdc159a1b4a0fa5883ddf562a696d00 As far as I remember this happened in some intermediate release.

src/test/java/javax/jmdns/impl/JmDNSTest.java
IvanBoychevMarinov: I just took a look and instances of the JmDNS are actually created few lines below in a for loop for each InetAddress.

mkorwel:
Hi 
@kaikreuzer. What next with this PR? Are you going to merge it? Are you planning next release with it?
@IvanBoychevMarinov thanks for you changes. I tested it in my case and your fix solves my problem.

kaikreuzer (approved):
Many thanks and sorry for the very late processing. Will do a new release before the end of the month!

PR #174, opened 2018-11-02
Remove waits. Fixes #144

### LINKED ISSUE #144: Registering Service is rather slow

cschneider:


kaikreuzer:
Although we are risking to have the implementation not fully respecting the specification anymore, I think we should merge it as this is a rather annoying behavior of the library for most users.

PR #183, opened 2018-12-17
Fixes #182 

### LINKED ISSUE #182: ServiceInfo which  named startwith ".","_" or endwith "." so that cannot be found
ServiceName like ".com“,"_x_" cannot be fond

zippozeng:


src/main/java/javax/jmdns/impl/ServiceInfoImpl.java
```diff
@@ -348,7 +348,6 @@ public ServiceInfoImpl(Map<Fields, String> qualifiedNameMap, int port, int weigh
             instance = "";
             // throw new IllegalArgumentException("The instance name component of a fully qualified service cannot be empty.");
         }
-        instance = removeSeparators(instance);
```
kaikreuzer: This line looks as if it was there for some important reason. Simply removing it might solve the one issue, but cause other regressions. Could you elaborate why you think it is not needed (or rather, did you understand why it was there in the first place)?

kaikreuzer:
Superseded by https://github.com/jmdns/jmdns/pull/179.

Issue #180, opened 2018-12-14
Allow escape of dot in instance name


mattiasbe:
Allow dots in instance names as per rfc 6763. According to rfc6763 instance names are allowed to contain any characters, including dots. They recommend escaping them if concatenating all the parts internally.

See section 4.3:
”4.3. Internal Handling of Names”
https://github.com/jmdns/jmdns/blob/master/devdocs/txt/rfc6763.txt

Also see this from avahi:
“Notes on DNS Name Escaping”
https://code.woboq.org/qt5/include/avahi-compat-libdns_sd/dns_sd.h.html

The reason for this requested support is that the European ITxPT public transport standard is using an additional extra field in the instance containing the separator dot.

Please see pull request #179 for a suggested minimum impact (?) version

kaikreuzer:
Fixed by https://github.com/jmdns/jmdns/pull/179.

axlsml:
I also created an issue according to contribution guidelines: https://github.com/jmdns/jmdns/issues/191

PR #190, opened 2019-06-17
Bug 191 Allow Dots in Instance Names for Subscription

axlsml:
Issue: https://github.com/jmdns/jmdns/issues/191
This is very related to https://github.com/jmdns/jmdns/issues/180 and https://github.com/jmdns/jmdns/pull/179.

The above changes by @mattiasbe fix the issue for **publishing** services However you cannot **subscribe** to such a service because the ServiceInfoImpl.decodeQualifiedNameMapForType() method can't handle the dot in instance names. 

Same reason here, I am quoting @mattiasbe:
> The reason for this requested support is that the European ITxPT public transport standard is using an additional extra field in the instance containing the separator dot.

For example subscribing to this won't work:
* instancename._apc._itxpt_http._tcp.local. :x: (instance should be instancename._apc)

However without the dot, it does:
* instancename_apc._itxpt_http._tcp.local. :heavy_check_mark: (instance should be instancename_apc)

So, in this pull request the decoding implementation is changed to work with the dot. I have removed the decode stuff from ServiceInfoImpl and put it into ServiceTypeDecoder. Also decode-related tests are moved from ServiceInfoTest to ServiceTypeDecoderTest. Plus a lot of tests are added, too.

---

Side note: I think also the decoding of subtypes is improved with this pull request. For example, the following type string was not decoded properly before:
* subtype._sub._apple-mobdev2._tcp.local. :x:

Apparently the previous implementation requires a leading underscore for subtypes. However RFC 6763, part "7.1.  Selective Instance Enumeration (Subtypes)" mentions:
> Subtype strings are not required to begin with an underscore, though
   they often do.  As with the TXT record key/value pairs, the list of
   possible subtypes, if any (including whether some or all begin with
   an underscore) are defined and specified separately for each basic
   service type.

Please, let me know what you think. Cheers! :)


axlsml:
I also created an issue according to the contribution guidelines: https://github.com/jmdns/jmdns/issues/191

axlsml:
@kaikreuzer please, any feedback here?

Can I do something to let travis-ci pass the build? I don't think its a problem related to those changes, is it?

rolfchess:
The travis-ci build fails because it wants to install java 7, but the installer script expects a version between 9 and 13:

```
install-jdk.sh 2019-05-02

Expected feature release number in range of 9 to 14, but got: 7

The command "~/bin/install-jdk.sh --target "/home/travis/openjdk7" --workspace "/home/travis/.cache/install-jdk" --feature "7" --license "GPL" --cacerts" failed and exited with 3 during .
``` 
I think that is part of the Travis environment and not a problem with the source code. I've built this pull request locally against Java 8 without problems.

@kaikreuzer can you help us out with integrating this pull request?

kaikreuzer:
Hey guys, sorry for having neglected that project over the past year...
Will try to catch up now. Just changed the travis config to use Java 8 instead of Java 7, which should make it work again.
Let me quickly close and reopen this PR to see, if it works.

kaikreuzer (approved):
Many thanks @axlsml for your thorough analysis, the detailed descriptions and the issue entry with a perfect PR for it 👍 .
Sorry again for having that not processed for such a long time 😞 . I'll do a new JmDNS release in October.
If you have any other bugs/fixes that you'd like to get in, let me know!

axlsml:
@kaikreuzer glad to hear. Thanks for merging! 👍

PR #197, opened 2020-02-20
Added a Thread.sleep in SocketListener#run method in order to allow o…

kolkoo:
…ther threads to get some CPU time if the current subnet is flooded with MDNS traffic.

The amount of time to sleep can be specified as a method argument of JmDNS#create method. If this argument is not used it defaults to 0 and no Thread.sleep calls is invoked.

Signed-off-by: Ivan Iliev <ivan.iliev@musala.com>


kaikreuzer (approved):
Thanks @kolkoo, looks good to me.
Sorry for not merging this any earlier. I hope you are doing fine!

PR #218, opened 2021-01-26
Allow overriding for JmDNS instance creation inside JmmDNSImpl

### LINKED ISSUE #215: Instanciating JmmDns is slow
I have had issues with the multihomed JmDns class where getting the instance of JmmDns took up to 30 seconds.

I have narrowed down the issue to the `JmDNS.create` calls to create new JmDns instance when the Network Topology changes, which is done multiple times when creating the JmmDns instance because it calls this function for every network interface.

In turn, creating those JmDns instances because they are created without specifing a name parameter to the `JmDns.create` function. When a name is not provided, the JmDns constructor will call `InetAddress.getHostName()` to find a suitable name for the instance.

It is known that calling `InetAddress.getHostName()` can take several seconds because it does a reverse hostname lookup. Doing so multiple times adds up to a significant amount of time.

I have done a small test program to test the calls to `JmDNS.create` with and without the name parameter. Here is the code and results for this:

```java
public class JmDNSSample
{
    public static void main(String[] args) throws IOException
    {
        long startTime, endTime;
        startTime = System.nanoTime();
        NetworkInterface.networkInterfaces()
            .flatMap(NetworkInterface::inetAddresses)
            .map(JmDNSSample::createInstanceWithHostNameLookup)
            .collect(Collectors.toList());

        endTime = System.nanoTime();

        System.out.println("Creating instances with hostname lookup took " + (endTime - startTime) / 1000000000.0 + "s");

        startTime = System.nanoTime();
        NetworkInterface.networkInterfaces()
            .flatMap(NetworkInterface::inetAddresses)
            .map(JmDNSSample::createInstanceWithoutHostNameLookup)
            .collect(Collectors.toList());

        endTime = System.nanoTime();

        System.out.println("Creating instances without hostname lookup took " + (endTime - startTime) / 1000000000.0 + "ns");
    }

    private static JmDNS createInstanceWithHostNameLookup(InetAddress addr)
    {
        try
        {
            return JmDNS.create(addr);
        }
        catch (IOException e)
        {
            e.printStackTrace();
        }

        return null;
    }

    private static JmDNS createInstanceWithoutHostNameLookup(InetAddress addr)
    {
        try
        {
            return JmDNS.create(addr, "HostName");
        }
        catch (IOException e)
        {
            e.printStackTrace();
        }

        return null;
    }
```

```
Creating instances with hostname lookup took 36.2592562s
Creating instances without hostname lookup took 0.5910996ns
```

We can see a significant improvement with a "hard-coded" name for `JmDNS.create` invocations. So, I suggest we had a way to specify a name that will be used in all calls to `JmDNS.create`. It can either be a `setName` function in JmmDNS, or we could add a function to ClassDelegate like `JmDNS createInstance(final InetAddress addr)` to let the user have more control over the creation of the JmDNS instances by the JmmDNS class, or any other solution that will allow me to skip the reverse hostname lookup when creating the JmmDNS instance.

For now as a workaround I can copy the implementation of JmmDNS and just change the offending line, but I would prefer we give more control to the user to create JmDNS instances used by JmmDNS.

NewLunarFire:
Resolves #215

You can replace the JmDNS instance creation by subclassing JmmDnsImpl

```java
import java.io.IOException;
import java.net.InetAddress;

import javax.jmdns.JmDNS;
import javax.jmdns.impl.JmmDNSImpl;

public class FastJmmDnsImpl extends JmmDNSImpl
{
    private static final String COMPUTER_NAME = System.getenv("COMPUTERNAME");

    @Override
    protected JmDNS createJmDnsInstance(InetAddress address) throws IOException
    {
        return JmDNS.create(address, COMPUTER_NAME);
    }
}
```

and then passing this class as the JmmDNS ClassDelegate:

```java
JmmDNS.Factory.setClassDelegate(FastJmmDnsImpl::new);
```

### DIFF
--- a/src/main/java/javax/jmdns/impl/JmmDNSImpl.java
+++ b/src/main/java/javax/jmdns/impl/JmmDNSImpl.java
@@ -629,7 +629,7 @@ public class JmmDNSImpl implements JmmDNS, NetworkTopologyListener, ServiceInfoI
             if (!_knownMDNS.containsKey(address)) {
                 synchronized (_knownMDNS) {
                     if (!_knownMDNS.containsKey(address)) {
-                        final JmDNS dns = JmDNS.create(address);
+                        final JmDNS dns = createJmDnsInstance(address);
                         if (_knownMDNS.putIfAbsent(address, dns) == null) {
                             // We need to register the services and listeners with the new JmDNS
                             final Collection<String> types = _serviceTypes;
@@ -783,6 +783,11 @@ public class JmmDNSImpl implements JmmDNS, NetworkTopologyListener, ServiceInfoI
             }
         }
 
+    }
+
+    protected JmDNS createJmDnsInstance(InetAddress address) throws IOException
+    {
+        return JmDNS.create(address);
     }
 
 }


piotrpo:
Hi, allowing for this change is essential. Overriding the way how the JmDNS is created is essential as currently there is no way to bind all interfaces to an own local domain.

kaikreuzer (approved):
Thank you!

Issue #211, opened 2020-10-31
Version 3.5.6 crash with java.util.ConcurrentModificationException in DNSCache


Tolriq:
It seems the rewrite now triggers java.util.ConcurrentModificationException: 

Unfortunately my stack trace does not embed line numbers due to R8 :( 

One of the getDNSEntry method is missing synchronization it seems.

```
java.util.ConcurrentModificationException: 
  at java.util.ArrayList$Itr.next (ArrayList.java:860)
  at javax.jmdns.impl.DNSCache.getDNSEntry (DNSCache.java)
  at javax.jmdns.impl.JmDNSImpl.handleRecord (JmDNSImpl.java)
                                getName (JmDNSImpl.java)
                                handleRecord (JmDNSImpl.java)
                      DNSEntry.isUnique (DNSEntry.java)
                      JmDNSImpl.handleRecord (JmDNSImpl.java)
                                getCache (JmDNSImpl.java)
                                handleRecord (JmDNSImpl.java)
                                getName (JmDNSImpl.java)
                                handleRecord (JmDNSImpl.java)
                                getCache (JmDNSImpl.java)
                                handleRecord (JmDNSImpl.java)
                      DNSRecord.getCreated (DNSRecord.java)
                      JmDNSImpl.isOlderThanOneSecond (JmDNSImpl.java)
                                handleRecord (JmDNSImpl.java)
                      DNSRecord.setWillExpireSoon (DNSRecord.java)
                      JmDNSImpl.handleRecord (JmDNSImpl.java)
                      DNSRecord.getTTL (DNSRecord.java)
                      JmDNSImpl.handleRecord (JmDNSImpl.java)
                      DNSRecord.setWillExpireSoon (DNSRecord.java)
                      JmDNSImpl.handleRecord (JmDNSImpl.java)
                                getCache (JmDNSImpl.java)
                                handleRecord (JmDNSImpl.java)
                      DNSEntry.sameSubtype (DNSEntry.java)
                      JmDNSImpl.handleRecord (JmDNSImpl.java)
                      DNSRecord.resetTTL (DNSRecord.java)
                      JmDNSImpl.handleRecord (JmDNSImpl.java)
                                getCache (JmDNSImpl.java)
                                handleRecord (JmDNSImpl.java)
                      DNSCache.replaceDNSEntry (DNSCache.java)
                      JmDNSImpl.handleRecord (JmDNSImpl.java)
                      DNSCache.replaceDNSEntry (DNSCache.java)
[...]

PR #227, opened 2021-05-17
Revert "Fixing desugaring issue"

kaikreuzer:
This reverts commit 6d747fb5dd9d36fc781d6dac056881b10585ff9c.

Fixes https://github.com/jmdns/jmdns/issues/211

### DIFF
--- a/src/main/java/javax/jmdns/impl/DNSCache.java
+++ b/src/main/java/javax/jmdns/impl/DNSCache.java
@@ -43,13 +43,11 @@ import org.slf4j.LoggerFactory;
  *
  * @author Arthur van Hoff, Werner Randelshofer, Rick Blair, Pierre Frisch
  */
-public class DNSCache {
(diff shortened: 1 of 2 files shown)


kaikreuzer:
@MWingbermuhle, @edgsousa and @HPSudip: Feel free to veto if you think this is a bad idea.

edgsousa:
What about the R8 issue? Is it still relevant?

The concurrent exceptions are most likely because lacking synchronized on the entrylist on the methods add & remove dns entries

kaikreuzer:
> What about the R8 issue? Is it still relevant?

According to https://github.com/jmdns/jmdns/issues/211#issuecomment-799595674, it does not seem to be relevant anymore.

kaikreuzer:
Ok, as nobody casts a veto, we all hope Google has fixed the issue on the Android side, so that it should be fine to revert this change and solve the NPEs and CMEs with it.

PR #228, opened 2021-05-17
Make mdns group address configurable

### LINKED ISSUE #222: Make mdns group address configurable
The MDNS_GROUP and MDNS_GROUP_IPV6 constants should be configurable, at least through a system property. The 224.0.0.0/24 subnet is link-only and would not be forwarded across subnets, same for the FF02 subnet in IPv6. Changing the address to be on the 239.0.0.0/24 (or FF04 in IPv6) would solve my issue and forward mdns packets across subnets.

So my suggestion is to make those addresses configurable with a system property and have the defaults be the addresses defined in the mdns spec. It can then be overriden using either `-Dnet.mdns.ipv4=239.0.0.251` as  a vm argument or `System.setProperty("net.mdns.ipv4", "239.0.0.251")` in Java code. The multicast port is already configurable this way.

```
public static final String MDNS_GROUP = System.getProperty("net.mdns.ipv4", "224.0.0.251");
public static final String MDNS_GROUP_IPV6 = System.getProperty("net.mdns.ipv6", "FF02::FB");
```

I will soon submit my patch as a PR.

NewLunarFire:
Fix #222

Allows configuration of mdns multicast group IP address by specifying them with the properties net.mdns.ipv4 and net.mdns.ipv6

### DIFF
--- a/src/main/java/javax/jmdns/impl/constants/DNSConstants.java
+++ b/src/main/java/javax/jmdns/impl/constants/DNSConstants.java
@@ -13,8 +13,8 @@ public final class DNSConstants {
     // http://www.iana.org/assignments/dns-parameters
 
     // changed to final class - jeffs
-    public static final String MDNS_GROUP                     = "224.0.0.251";
-    public static final String MDNS_GROUP_IPV6                = "FF02::FB";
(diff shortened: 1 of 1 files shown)


kaikreuzer (approved):
lgtm, thank you!

PR #235, opened 2021-11-23
getServiceInfo returns null sometimes

zhangl22000:
JmDNS instance invokes getServiceInfo,  but did not wait for enough time, then returned null value.  

When the ServiceInfo's updateRecord method was invoked, even if there is not any data change in this ServiceInfo,  It will notify all the threads which wait  at the ServiceInfo.  That's why the  wait method in waitForInfoData was awaken unexpected.

The correct behavior is:   The ServiceInfo will not notifyAll if there is not any data change.  If there is data change, the ServiceInfo will notify that it is changed.  Then the info.wait(200) in waitForInfoData will be notified.

### DIFF
--- a/src/main/java/javax/jmdns/impl/ServiceInfoImpl.java
+++ b/src/main/java/javax/jmdns/impl/ServiceInfoImpl.java
@@ -764,14 +764,13 @@ public class ServiceInfoImpl extends ServiceInfo implements DNSListener, DNSStat
                     ServiceEvent event = new ServiceEventImpl(dns, this.getType(), this.getName(), this.clone());
                     dns.handleServiceResolved(event);
                 }
+                // This is done, to notify the wait loop in method JmDNS.waitForInfoData(ServiceInfo info, int timeout);
+                synchronized (this) {
+                    this.notifyAll();
+                }
             } else {
                 logger.debug("JmDNS not available.");
             }
-        }
-
-        // This is done, to notify the wait loop in method JmDNS.waitForInfoData(ServiceInfo info, int timeout);
-        synchronized (this) {
-            this.notifyAll();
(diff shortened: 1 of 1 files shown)


src/main/java/javax/jmdns/impl/ServiceInfoImpl.java
```diff
@@ -764,15 +764,14 @@ public void updateRecord(final DNSCache dnsCache, final long now, final DNSEntry
...
+                }
             } else {
                 logger.debug("JmDNS not available.");
             }
         }
-
-        // This is done, to notify the wait loop in method JmDNS.waitForInfoData(ServiceInfo info, int timeout);
-        synchronized (this) {
-            this.notifyAll();
-        }
```
qingfeng0820, line 775: I think if change the notifyAll() here may cause problem:
If notifyAll() when service info changed but the service info has not changed any more, it may be no chance to wait up the waiters (without wait timeout) of service info then. 

I think It's better to change the JmDNSImpl.waitForInfoData to wait enough time instead of changing the notifyAll() here.

qingfeng0820: I checked with @zhangl22000, the service info needs to call notifyAll when the dnsEntry is related this service info.

zhangl22000: Both fixes are fine. Fix option 1: when service info has changes, notifyAll() is called.  As waiForInfoData() only wait when service info has no data). Fix option 2: notifyAll() when this service info is updated  (updated, but may not change).

kaikreuzer (approved):
Thanks for the fix!

PR #240, opened 2022-03-18
Always perform NetworkChecker task in background

wallisch:
On modern macOS versions, the `NetworkChecker` can take a long time to finish.

This lets the Timer perform the check in background immediately after invocation, instead of blocking the Initializers by calling `run()` directly.


kaikreuzer (approved):
lgtm, thank you.
I hope that there wasn't any specific reason why the first execution was done synchronously...

PR #242, opened 2022-08-18
Taking the version of SLF4J-API to latest available at this moment.

MWingbermuhle:
Version 1.7.25 is already 3 years old, and Synopsis Detect is giving operational risk on it.


kaikreuzer (approved):
Thanks!

PR #247, opened 2023-01-15
Slightly reduced memory usage of the DNSCache

joerg1985:
Use more accurate map implementations and presize the lists inside the 
cache to reduce the memory usage.

### DIFF
--- a/src/main/java/javax/jmdns/impl/DNSCache.java
+++ b/src/main/java/javax/jmdns/impl/DNSCache.java
@@ -210,7 +210,7 @@ public class DNSCache extends ConcurrentHashMap<String, List<DNSEntry>> {
         if (dnsEntry != null) {
             List<DNSEntry> entryList = this.get(dnsEntry.getKey());
             if (entryList == null) {
-                this.putIfAbsent(dnsEntry.getKey(), new ArrayList<DNSEntry>());
+                this.putIfAbsent(dnsEntry.getKey(), new ArrayList<DNSEntry>(3));
                 entryList = this.get(dnsEntry.getKey());
             }
             synchronized (entryList) {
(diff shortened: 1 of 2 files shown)


joerg1985:
Here are details to this PR, i am running openHab 3.4 on a Raspi, which is a memory constrained environment.
Sometimes the JmDNSImpl is one of the object with the biggest memory footprint.

image

Most of this is due to the _cache field
image

A significant part of each entry in the cache is the _qualifiedNameMap, which is replaced in this PR.
image

joerg1985:
@kaikreuzer could you have a look at this? thanks in advance

kaikreuzer:
@joerg1985 Interesting change, thank you! It is actually the first time I hear about `EnumMap` - sounds useful indeed.
But will this change have any significant impact? As you showed, the cache size is somehow much bigger than the value it brings. Should we think about any further measures, like e.g. converting the cache to a LRU cache?

MWingbermuhle (approved):
Good improvements, with not much risk. 

According to the Javadoc of ArrayList, the parameter of the constructor is the initialCapacity, and it will grow the capacity when needed, but this indeed will initially reduce the memory footprint of the cache, but not actually limit it. The only concern here is that growing the capacity will cost performance because of allocation times on constrained environments. If that poses a problem to somebody at some point, he/she can consider filing a PR to make that configurable somewhere.

Using the EnumMap also seems like a good alternative, and will not pose a risk at all. The semantics and behavior of EnumMap and HashMap are identical when using a Enum as key.

kaikreuzer (approved):
Thanks for the PR, @joerg1985 and thanks for the review @MWingbermuhle!

Let's merge this then and have it as part of the next release.

PR #253, opened 2023-07-19
Improve service discovery

carlonzo:
During my test discovering few services published from the official android mdns implementation, I found few issue that I tried to debug and resolve here. After those fixes, jmdns can reliably discover all the services, when before only few of them where discovered, and not even immediately.

I am opening a PR with the fixes I applied hoping to start a discussion about it.

- `decodeQualifiedNameMapForType` was not able to parse correctly the A/AAA records because the published type is of the format `android_1234.local.`. this was breaking as the `_qualifiedNameMap` recognized the type as only `application` name. Checking the RFC and browsing I came to the conclusion that A/AAA records can only be formatted as `<instance>.<domain>`. My fix simply allows to specify what record we are decoding. This solved the majority of the issues where the A/AAA record where stored in cache with the wrong key and never found

-  `getServiceInfo()` for a `SRV DNSRecord` is not properly setting the `_server` field but instead is built with an empty string. This was breaking in my test when the cache tried to resolve the serviceinfo from a SRV DNSEntry from cache returning an empty server string (when instead the server field was available in the DNSEntry)

Again I am more then happy to discuss my changes and apply a better fix as you see fit


src/main/java/javax/jmdns/impl/ServiceTypeDecoder.java
```diff
@@ -40,7 +45,7 @@ static Map<Fields, String> decodeQualifiedNameMapForType(String type) {
             name = ServiceInfoImpl.removeSeparators(casePreservedType.substring(0, index));
             domain = casePreservedType.substring(index);
             application = "";
-        } else if ((!aType.contains("_")) && aType.contains(".")) {
+        } else if (recordType == DNSRecordType.TYPE_A || recordType == DNSRecordType.TYPE_AAAA || (!aType.contains("_")) && aType.contains(".")) {
```
MWingbermuhle, line 48: There should be no risk here, but I like to see these added checks inverted (like `DNSRecordType.TYPE_A == recordType`) because of the nullability of recordType. If anyone ever tries to convert this to a .equals operation in the future, it will not cause a NullPointerException that way.

MWingbermuhle:
@kaikreuzer asked me to review, so here I go.

src/main/java/javax/jmdns/impl/DNSEntry.java
```diff
@@ -45,12 +45,16 @@ public abstract class DNSEntry {
...
         _dnsClass = recordClass;
         _unique = unique;
-        _qualifiedNameMap = ServiceTypeDecoder.decodeQualifiedNameMapForType(this.getName());
+        _qualifiedNameMap = ServiceTypeDecoder.decodeQualifiedNameMapForType(this.getName(), type);
         String domain = _qualifiedNameMap.get(Fields.Domain);
         String protocol = _qualifiedNameMap.get(Fields.Protocol);
         String application = _qualifiedNameMap.get(Fields.Application);
         String instance = _qualifiedNameMap.get(Fields.Instance).toLowerCase();
-        _type = (application.length() > 0 ? "_" + application + "." : "") + (protocol.length() > 0 ? "_" + protocol + "." : "") + domain + ".";
+        String _type = (application.length() > 0 ? "_" + application + "." : "") + (protocol.length() > 0 ? "_" + protocol + "." : "") + (domain.length() > 0 ? domain + "." : "");
```
MWingbermuhle, line 53: Firstly, I'm missing unittests for this change. 
Next to that, without Null-checks, the `length()` operations are dangerous and will cause NPE's at some point. 

Since the extra Null-checks will cause this piece of code to grow, you might consider extracting a method out of this. That will also benefit testability.

src/main/java/javax/jmdns/impl/DNSRecord.java
```diff
@@ -861,7 +861,7 @@ DNSOutgoing addAnswer(JmDNSImpl dns, DNSIncoming in, InetAddress addr, int port,
          */
         @Override
         public ServiceInfo getServiceInfo(boolean persistent) {
-            return new ServiceInfoImpl(this.getQualifiedNameMap(), _port, _weight, _priority, persistent, (byte[]) null);
+            return new ServiceInfoImpl(this.getQualifiedNameMap(), _port, _weight, _priority, persistent, _server);
```
MWingbermuhle, line 864: Small change, but again no test?

src/test/java/javax/jmdns/impl/ServiceTypeDecoderTest.java
```diff
@@ -217,7 +218,7 @@ public void testReverseDNSQuery() {
     public void testAddress() {
         String type = "panoramix.local.";
 
-        Map<ServiceInfo.Fields, String> map = ServiceTypeDecoder.decodeQualifiedNameMapForType(type);
```
MWingbermuhle, line 220: Since you made the code use the OR operator to allow also the new recordType, this test became valid in more than 1 way. So please create a  new test cases that tests the TYPE_A and Type_AAA methods, but leave this case intact with the original method call.

MWingbermuhle (changes requested):
Good ideas, but not quite there yet.

carlonzo:
Thanks @MWingbermuhle for the review. I agree with your points and I'll apply the suggestions.
About tests obviously I agree with you, but I was not sure if the proposed changes would have been accepted. I'll get them ready right away

src/main/java/javax/jmdns/impl/DNSEntry.java
kaikreuzer, line 53: The (not very Java-like) naming convention in this code seems to be to have fields starting with a `_`. As you are now converting this to a local variable, I guess it would be better to rename the variable to `type`.

src/test/java/javax/jmdns/impl/ServiceTypeDecoderTest.java
carlonzo: change removed

src/main/java/javax/jmdns/impl/DNSRecord.java
carlonzo: done thanks

src/main/java/javax/jmdns/impl/DNSEntry.java
carlonzo: done thanks!

carlonzo: done! thanks

src/test/java/javax/jmdns/test/DNSRecordTest.java
```diff
@@ -99,4 +100,18 @@ public void testIPv4MappedIPv6Addresses() throws Exception {
...
+
+        DNSRecord record = new DNSRecord.Service("test._http._tcp.local.", DNSRecordClass.CLASS_IN, true, TTL_IN_SECONDS, 0, 0, 0, "test_server");
+
+        ServiceInfo serviceInfo = record.getServiceInfo(true);
+
+        Assert.assertEquals("test", serviceInfo.getName());
+        Assert.assertEquals("http", serviceInfo.getApplication());
+        Assert.assertEquals("tcp", serviceInfo.getProtocol());
+        Assert.assertEquals("local", serviceInfo.getDomain());
+        Assert.assertEquals("test_server", serviceInfo.getServer());
```
carlonzo, line 115: without my fix, `serviceInfo.getServer()` was returning an empty string when instead it was part of the original `record`

carlonzo:
Sorry for the delay applying the changes required. Had personal impediments and did not find the time to think about a better approach. I removed one of the fixes from this PR and moved to https://github.com/jmdns/jmdns/pull/261 where I think it requires a separate discussion

src/test/java/javax/jmdns/test/DNSRecordTest.java
```diff
@@ -99,4 +100,30 @@ public void testIPv4MappedIPv6Addresses() throws Exception {
...
+    }
+
+    @Test
+    public void testDNSRecordKey() {
+        DNSRecord record1 = new DNSRecord.Service("android_123.local.", DNSRecordClass.CLASS_IN, true, TTL_IN_SECONDS, 0, 0, 0, "test_server");
+        Assert.assertEquals("_android_123.local.", record1.getKey());
```
carlonzo, line 121: before the fix, the key was `_android_123.local..` (double dot at the end)

kaikreuzer (approved):
lgtm, many thanks for these fixes!

Issue #13, opened 2014-10-22
Checking of instance name lengths


jazdw:
It doesn't seem like JmDNS does any instance name length checking. According to RFC 6763 the instance name should be limited to 63 octets after conversion to UTF-8.

Its not something that should be handled in client code as JmDNS can append numbers the instance name when there is a duplicate instance name on the network.

melloware:
@kaikreuzer this was fixed by: https://github.com/jmdns/jmdns/pull/255 and can be closed

ViToni:
Actually #255 doesn't solve this issue. The OP correctly mentions
> name should be limited to 63 octets after conversion to UTF-8.

and
> as JmDNS can append numbers the instance name when there is a duplicate instance name on the network

#255 doesn't handle conversion nor a suffix in case of a name collision.

melloware:
@ViToni it sounds like there are 2 different issues in this one ticket

- [ ] The 63 octet limit
- [ ] The collision check

PR #255, opened 2023-08-10
Ensure DNS label cannot exceed 63 characters in length

LarsGjesse-Zoetis:


src/main/java/javax/jmdns/impl/HostInfo.java
```diff
@@ -40,6 +40,8 @@ public class HostInfo implements DNSStatefulObject {
 
     private final HostInfoState _state;
 
+    private final static int    _labelLengthLimit = 0x3F;
```
kaikreuzer, line 43: Is this a constraint defined in the mDNS spec?

src/main/java/javax/jmdns/impl/HostInfo.java
```diff
@@ -104,6 +106,11 @@ public static HostInfo newHostInfo(InetAddress address, JmDNSImpl dns, String jm
         if (index > 0) {
             aName = aName.substring(0, index);
         }
+        if (aName.length() > _labelLengthLimit) {
```
kaikreuzer, line 109: Can we assume that if this situation happens, the discovered device does not correctly respect the specification?
If so, we should add at least some logging (warn or maybe better info level?).

src/main/java/javax/jmdns/impl/HostInfo.java
```diff
@@ -104,6 +106,11 @@ public static HostInfo newHostInfo(InetAddress address, JmDNSImpl dns, String jm
...
+            // Remove trailing labels which would make the combined label exceed 63 characters in length
```
kaikreuzer, line 110: Could you maybe add an example in the comment on what a too-long-label would be shortened to? (i.e. show why to better cut it before the last "." instead of simply limiting it to 63 chars.

src/main/java/javax/jmdns/impl/HostInfo.java
LarsGjesse-Zoetis: It's a constraint of https://www.ietf.org/rfc/rfc1035.txt, section 2.3.4

src/main/java/javax/jmdns/impl/HostInfo.java
LarsGjesse-Zoetis: I don't think the device violates the spec. It's more the way that JmDNS converts an interface name to a single DNS label.
Concretely, I saw this problem in a situation where I had JmDNS running in an Azure pipeline job.
The Azure VM's network interface name is something like fv-az767-139.5kzmkpjakaaulcrpfkoga5aama.gx.internal.cloudapp.net. When the dots are replaced by dashed, it becomes a single DNS label which exceeds 63 characters. In the original interface name, each segment is less than 63 characters, and the whole name is less than 255 characters.

src/main/java/javax/jmdns/impl/HostInfo.java
LarsGjesse-Zoetis: It could have been done in other ways, I agree. It was just what seemed most natural to me.
Following the example above, the resulting name would then be fv-az767-139-5kzmkpjakaaulcrpfkoga5aama-gx-internal-cloudapp.local

src/main/java/javax/jmdns/impl/HostInfo.java
kaikreuzer: Perfect, thanks.

kaikreuzer (approved):
Many thanks for your explanations and the fix!

melloware:
I am pretty this closes: https://github.com/jmdns/jmdns/issues/13

src/main/java/javax/jmdns/impl/HostInfo.java
ViToni: The DNS spec ([RFC 1035 - Section 2.3.1](https://www.ietf.org/rfc/rfc1035.html#section-2.3.1)) is based on plain text. In this case every character would be an octet.
mDNS however is based on UTF-8 (see [Multicast DNS Character Set](https://datatracker.ietf.org/doc/html/rfc6762#section-16)).
As the maximum number of octets per label has not changed one would need to check the number of bytes needed to encode a given UTF-8 string against the limit of 63 octets.
Cutting off based on the limit might be also more complex because "simply" cutting of after the limit could cut through an encoded UTF-8 character making the resulting label invalid UTF-8.

ViToni (commented):
First of all, sorry for being late to the game. I hadn't seen this change before.

PR #261, opened 2023-11-18
Improve support to atype service type

carlonzo:
Moving here fix from https://github.com/jmdns/jmdns/pull/253

`decodeQualifiedNameMapForType` was not able to parse correctly the A/AAA records because the published type is of the format `android_1234.local.`. This was breaking as the `_qualifiedNameMap` recognized the type as only application name due to the underscore in the instance name. Checking the RFC and browsing I came to the conclusion that A/AAA records can only be formatted as <instance>.<domain>. The fix adds an additional regex to support the A/AAA record type format. This solved the majority of the issues where the A/AAA record where stored in cache with the wrong key and never found.

Big problem I encountered and probably will create an issue here is that one of the behaviors changed, and so the test.

 One of the tests `decode` parses the type `DIST123_7-F07_OC030_05_03941.local.` as a lowercase `application`. This goes in contrast with all other test cases where the case of the type are all preserved and the type was not parsed and falled back to one `application`.
 
 one of the other tests like  `testAddress` instead parses and split`panoramix.local.` into 2 parts: `panoramix` (instance/name) and `local` (domain) maintaining the case. The 2 types have the same structure but the underscore which is an allowed character. 
 
 Open to start a discussion about this, but I clearly see a case where the checks and tests are contrasting.

### DIFF
--- a/src/main/java/javax/jmdns/impl/ServiceTypeDecoder.java
+++ b/src/main/java/javax/jmdns/impl/ServiceTypeDecoder.java
@@ -11,6 +11,8 @@ class ServiceTypeDecoder {
     private static final Pattern SUBTYPE_PATTERN = Pattern.compile("^((.*)\\._)?_?(.*)\\._sub\\._([^.]*)\\._([^.]*)\\.(.*)\\.?$");
     private static final Pattern PATTERN = Pattern.compile("^((.*)?\\._)?([^.]*)\\._([^.]*)\\.(.*)\\.?$");
 
+    private static final Pattern TYPE_A_PATTERN = Pattern.compile("^([^.]*)\\.(.*)\\.?$");
+
     private ServiceTypeDecoder() {
     }
 
@@ -40,11 +42,6 @@ class ServiceTypeDecoder {
             name = ServiceInfoImpl.removeSeparators(casePreservedType.substring(0, index));
             domain = casePreservedType.substring(index);
             application = "";
-        } else if ((!aType.contains("_")) && aType.contains(".")) {
-            index = aType.indexOf('.');
-            name = ServiceInfoImpl.removeSeparators(casePreservedType.substring(0, index));
-            domain = ServiceInfoImpl.removeSeparators(casePreservedType.substring(index));
-            application = "";
         } else {
             Matcher subType = SUBTYPE_PATTERN.matcher(aType);
             if (subType.matches()) {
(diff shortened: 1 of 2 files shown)


src/main/java/javax/jmdns/impl/ServiceTypeDecoder.java
```diff
@@ -40,11 +42,6 @@ static Map<Fields, String> decodeQualifiedNameMapForType(String type) {
             name = ServiceInfoImpl.removeSeparators(casePreservedType.substring(0, index));
             domain = casePreservedType.substring(index);
             application = "";
-        } else if ((!aType.contains("_")) && aType.contains(".")) {
```
carlonzo, line 43: this part is covered now by the new regex

src/test/java/javax/jmdns/impl/ServiceTypeDecoderTest.java
```diff
@@ -34,7 +34,7 @@ public void decodeWithSubtype2() {
     @Test
     public void decode() {
         Map<ServiceInfo.Fields, String> actual = ServiceTypeDecoder.decodeQualifiedNameMapForType("DIST123_7-F07_OC030_05_03941.local.");
-        Map<ServiceInfo.Fields, String> expected = ServiceInfoImpl.createQualifiedMap("", "dist123_7-f07_oc030_05_03941.local", "", "", "");
+        Map<ServiceInfo.Fields, String> expected = ServiceInfoImpl.createQualifiedMap("DIST123_7-F07_OC030_05_03941", "", "", "local", "");
```
carlonzo, line 37: case is preserved when resolving

carlonzo:
> I'm not sure I can fully grasp the remaining issue here. Your fix seems to make sense if the RfC allows underscores in A/AAA records (which seems to be the case).
> 
> > One of the tests decode parses the type `DIST123_7-F07_OC030_05_03941.local.` as a lowercase application. 
> 
> So do you consider this test to be correct or is it simply a crappy test that we should ignore/delete from your pov?
> 

IMHO a wrong test given the underscore and the fact that only allow lowercase when everything else maintains cases

carlonzo:
The test we can keep it but it needs to be modified.

kaikreuzer:
Ok, so then your changes should all be fine and there isn't any need for discussion then - even better!

kaikreuzer (approved):

carlonzo:
oh wow! thanks!

Issue #246, opened 2022-12-09
Rare StringIndexOutOfBoundsException in ServiceTypeDecoder.originalCase()


bubbleguuum:
Seen this exception in an automated crash report for my program:

```
java.lang.StringIndexOutOfBoundsException: begin 48, end 54, length 53"
    at java.base/java.lang.String.checkBoundsBeginEnd(Unknown Source)
    at java.base/java.lang.String.substring(Unknown Source)
    at javax.jmdns.impl.ServiceTypeDecoder.originalCase(ServiceTypeDecoder.java:73)
    at javax.jmdns.impl.ServiceTypeDecoder.decodeQualifiedNameMapForType(ServiceTypeDecoder.java:62)
    at javax.jmdns.impl.DNSRecord$Pointer.getServiceInfo(DNSRecord.java:544)
    at javax.jmdns.impl.DNSRecord$Pointer.getServiceEvent(DNSRecord.java:555)
    at javax.jmdns.impl.JmDNSImpl.updateRecord(JmDNSImpl.java:1287)
    at javax.jmdns.impl.JmDNSImpl.handleRecord(JmDNSImpl.java:1447)
    at javax.jmdns.impl.JmDNSImpl.handleQuery(JmDNSImpl.java:1569)
    at javax.jmdns.impl.SocketListener.run(SocketListener.java:75)

```

Looking at `ServiceTypeDecoder`, crashing line is:

```
    return casePreservedType.substring(matcher.start(group), matcher.end(group));
```

For some reason, `matcher.end(group)` (with group=5) is returning an index value in the input string of the `Matcher`
greater than the string length (54 > 53). Maybe a funky `type` parameter to `decodeQualifiedNameMapForType()` with unicode chars or something ?

JRE is Eclipse Adoptium 11.0.13+8.

ItsAMeNick:
I've been seeing this as well.

`java.lang.StringIndexOutOfBoundsException: begin 21, end 27, length 26
    at java.lang.String.checkBoundsBeginEnd(String.java:3942)
    at java.lang.String.substring(String.java:2467)
    at javax.jmdns.impl.ServiceTypeDecoder.originalCase(ServiceTypeDecoder.java:16)
    at javax.jmdns.impl.ServiceTypeDecoder.decodeQualifiedNameMapForType(ServiceTypeDecoder.java:130)
    at javax.jmdns.impl.DNSRecord$Pointer.getServiceInfo(DNSRecord.java:75)
[...]

PR #269, opened 2024-02-06
Do not Pattern match lowerCased ServiceType not to get IndexOutOfBoun…

plotczyktest:
Do not Pattern match lowerCased ServiceType not to get IndexOutOfBoundsException while trying extracting casePreservedType matched groups containing special characters.

Lower casing type is only valid for "arpa" types mdns records, where we compare texts "in-addr.arpa" or "ip6.arpa"

For other types we do not need to work on lowerCased type so we can directly substring params based on casePreservedType Matched groups for given type

fixes([246](https://github.com/jmdns/jmdns/issues/246))


kaikreuzer (approved):
Many thanks for your analysis and the fix!

PR #201, opened 2020-04-16
Remove entries with no records (jmdns#200)

edgsousa:
Signed-off-by: Edgar Sousa <edgar.sousa@tridonic.com> (edgsousa)

Issue #200, opened 2020-04-16
DNSCache leak


edgsousa:
Current `DNSCache` implementation does not remove hashmap entries after all `DnsEntry` objects are removed for that key.

In a long running system, this leads to OutOfMemory errors after a couple hundred thousands of different hostnames and services were present at some moment on the network.

kaikreuzer:
Fixed with https://github.com/jmdns/jmdns/pull/201.

Issue #186, opened 2019-04-04
OOM and high CPU caused by door camera


ivenhov:
Hi all

I have a strange problem caused by our door bell camera discovered through mDNS
Problem eventually leads to 100% CPU and OutOfMemory errors on my JVM.

Camera sends packet which looks like this

`
000084000000000e00000000184772616e6473747265616d2d303030423832423834364434056c6f63616c00000180010000007800040b0100320235300130013102313107696e2d6164647204617261700000018001000000780002c00c224772616e6473747265616d2047445333373130202d20303030423832423834364434055f68747470045f746370056c6f63616c00002180010000007800080000000001bbc00cc05e0010800100001194000100095f7365727669636573075f646e732d7364045f756470056c6f63616c00000c0001000011940002c081c081000c0001000011940002c05e224772616e6473747265616d2047445333373130202d20303030423832423834364434055f72747370045f746370056c6f63616c000021800100000078000800000000022ac00cc0ea0010800100001194000100c0b2000c0001000011940002c0eac0ea000c0001000011940002c0ea224772616e6473747265616d2047445333373130202d20303030423832423834364434125f4772616e6473747265616d2d766964656f045f746370056c6f63616c00002180010000007800080000000001bbc00cc15a0010800100001194000100c0b2000c0001000011940002c0eac0ea000c0001000011940002c15a
`

This can be broken down as follows (bulk at the end omitted)

```
Some header
000084000000000e00000000 

First Type A answer
18 4772616e6473747265616d2d303030423832423834364434056c6f63616c00000180010000007800040b010032

2nd Type A record
02 35300130013102313107696e2d61646472046172617000000180010000007800 02 c00c
```
I have no problem with first answer
`[IPv4Address@1341712829 type: TYPE_A index 1, class: CLASS_IN index 1-unique, name: Grandstream-000B82B846D4.local. ttl: '0/120' address: '11.1.0.50']`

But the 2nd one is funny. Last part 02 c00c indicates there are 2 bytes for records data, and that really should be 4 bytes as far as I can tell as this is Inet4Address

This looks like a bug in the camera. We've upgraded to the latest firmware but the problem stays. We will be in touch with manufacturer as well.

The thing is JmDNS still creates record but with IP which is null.
```
java.net.UnknownHostException: addr is of illegal length
	at java.net.InetAddress.getByAddress(InetAddress.java:1042)
	at java.net.InetAddress.getByAddress(InetAddress.java:1439)
[...]

PR #271, opened 2024-02-08
Fix for OOM and concurrency issues in DNSCache 

ivenhov:
Hi

The changes are two address 2 issues

problems with door bell camera malformed records, see OOM and high CPU caused by door camera #186
concurrency issues in DNSCache
The problem has been reported here DNSCache leak #200
Fix is provided by Remove entries with no records (jmdns#200) #201 and included in 3.5.6.
Diff for the fix https://github.com/jmdns/jmdns/pull/201/files
It seems however that the fix is incomplete and will have other issues

concurrency issue on entryList.isEmpty() method
concurrency issue with record being added whilst entryList is removed at the same time
This change provides more coarse locking to solve above issues.

### DIFF
--- a/src/main/java/javax/jmdns/impl/util/SimpleLockManager.java
+++ b/src/main/java/javax/jmdns/impl/util/SimpleLockManager.java
@@ -1,0 +1,72 @@
+/*
+ * Copyright Object Matrix 2017
+ */
+
+package javax.jmdns.impl.util;
+
+import java.io.Closeable;
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.TimeUnit;
+
+/**
+ * Basic lock manager which uses internal {@link Map} to maintain a list of current locks.
+ */
+public class SimpleLockManager {
+
+    private final ConcurrentHashMap<String, Locked> _locks = new ConcurrentHashMap<>();
+
+    public Locked lock(String lockKey) {
+        try {
+            return tryLock(lockKey, Long.MAX_VALUE, TimeUnit.MILLISECONDS);
+        } catch (LockFailedException e) {
+            throw new RuntimeException(e);
+        } catch (InterruptedException e) {
+            Thread.currentThread().interrupt();
+            throw new RuntimeException(e);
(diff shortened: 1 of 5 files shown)


src/main/java/javax/jmdns/impl/util/SimpleLockManager.java
```diff
@@ -0,0 +1,72 @@
...
+        } catch (LockFailedException e) {
+            throw new RuntimeException(e);
+        } catch (InterruptedException e) {
+            Thread.currentThread().interrupt();
+            throw new RuntimeException(e);
+        }
+    }
+
+    @SuppressWarnings("resource")
+    public Locked tryLock(String lockKey, long time, TimeUnit timeunit) throws InterruptedException, LockFailedException {
```
kaikreuzer, line 31: Instead of implementing the mechanism ourselves, I would prefer to use standard Java mechanisms such as `Semaphore`. Or am I missing any reason why this isn't possible here?

kaikreuzer (changes requested):
Thanks for this follow-up fix.
I have one review comment - please see below.

src/main/java/javax/jmdns/impl/util/SimpleLockManager.java
ivenhov: Hi
Thanks for your feedback

My idea was as follows
- reduce contention to a specific key, rather than a global lock
- provide Lock implementing  Closeable interface to use in try-with-resource and have early compilation warning if not closed. Instead Java's Lock interface would work here

Semaphore is usually used to control number of concurrent threads. Here likely it would be used with number of permits = 1, so a global lock, equivalent to synchronized block
It would still require using something like ConcurrentHashMap for keeping track of key->lock mapping

I'm happy to re-implement it if you can think of a simple pattern I could use here?

kaikreuzer: Sorry for dropping the ball here - thanks for your explanations!

kaikreuzer (approved):
lgtm, thanks!

PR #278, opened 2024-07-21
Collect the name inside a single StringBuilder

joerg1985:
This PR will collect the name inside a single `StringBuilder` and create the substrings at the end, instead of maintaining the substrings while reading each fragment of the name.

Signed-off-by: Jörg Sautter <joerg.sautter@gmx.net>


kaikreuzer (approved):
lgtm, thanks!

Issue #281, opened 2024-08-07
Update 3.5.10 and 3.5.11 brokes ServiceEvent in discovery


GazimSoliev:
ServiceEvent in discovery gives wrong info about service, for example empty addresses (ServiceEvent.info.hostAddresses) and wrong hostname from field "ServiceEvent.info.server"

kaikreuzer:
@Siedlerchr, @ivenhov, @plotczyktest & @joerg1985: You are the contributors of the changes to these releases - does any of you have an idea, what change could cause this regression and how to fix it?

kaikreuzer:
@GazimSoliev Can you maybe provide some more details on the issue like e.g. a concrete example of what is "wrong"?

Siedlerchr:
Without any more details it's hard to judge here what is wrong or not. 
My changes did not affect the content of the ServiceInfo message itself, and I am using it for an IPP printer server without any problems

GazimSoliev:
Tested in Kotlin Multiplatform 2.0.0, target JVM and Android JVM. In this test I caught that JmDNS can't resolve service in version 3.5.11, but in 3.5.9 everything works.
dependecies:
org.jmdns:jmdns:3.5.11
org.jetbrains.kotlinx:kotlinx-coroutines-core:1.9.0-RC
org.jetbrains.kotlin:kotlin-test:2.0.0

```
import kotlinx.coroutines.CompletableDeferred
import kotlinx.coroutines.runBlocking
import kotlinx.coroutines.withTimeout
import java.net.Inet4Address
import java.net.Inet6Address
import java.net.InetAddress
import java.net.NetworkInterface
import javax.jmdns.JmDNS
import javax.jmdns.ServiceEvent
import javax.jmdns.ServiceInfo
import javax.jmdns.ServiceListener
import kotlin.test.*

class CheckJmDNS {
    private val iPv4Address = getIPAddressInLan()
    private val hostname = "hostname-test"
    private val type = "_http._tcp.local."
    private val name = "example"
    private val port = 1234
    private lateinit var jmdns: JmDNS
    private val deferredInfo = CompletableDeferred<ServiceInfo>()

    @BeforeTest
    fun createJmdns() {
        jmdns = JmDNS.create(iPv4Address, hostname)
        val serviceInfo = ServiceInfo.create(type, name, port, "path=index.html")
        val listener = object : ServiceListener {
            override fun serviceAdded(event: ServiceEvent?) {
                println("Added service: $event")
            }

[...]

PR #275, opened 2024-06-20
Create less temporary strings

joerg1985:
This PR will use a `StringBuilder` in areas where a lot of temporary strings are created.
To lower the object fallout while handling DNS requests. 

Signed-off-by: Jörg Sautter <joerg.sautter@gmx.net>

PR #283, opened 2024-08-22
Fix #291 by removing the faulty domain prefix

joerg1985:
This is the fix to the issue #281 introduced in #275, sorry for breaking it.

Signed-off-by: Jörg Sautter <joerg.sautter@gmx.net>


ViToni (approved):
LGTM

kaikreuzer (approved):
Thanks for fixing!

Issue #229, opened 2021-06-08
Overly extensive and superfluous Logging in DNSIncoming.readAnswer


geroyche:
I see my log being spammed.
There are sort of two issues here:

1. how can logging be held to an acceptable level
2. what triggers the error

Regarding 1.:
in my case, the following condition is true, and I get a substantial log entry, as I have 320 answers in that array:
https://github.com/jmdns/jmdns/blob/main/src/main/java/javax/jmdns/impl/DNSIncoming.java#L312

then, TYPE_UNKNOWN is not handled in the subsequent condition here:
https://github.com/jmdns/jmdns/blob/main/src/main/java/javax/jmdns/impl/DNSIncoming.java#L317
which leads to what's essentially a second full printout of 320 answers.

I would submit a pull request, but I am not deep enough into the code yet to understand what would be the best solution... early returns? a simple && (type != DNSRecordType.TYPE_UNKNOWN) ?

also: I question whether print() should actually do all that it does when the application is not deliberately set to debug mode?

Regarding 2:
The parser seems to take issues with this packet, and I am not sure why, but I am also not really firm in DNS payloads.
Here's the start of the packet dump:

` 0: 00008400000001a9 000000000f4c4150 544f502d49363238 35434d54056c6f63 `
aka
`........ .....LAP TOP-I628 5CMT.local......`

apparently the parser considers this a compressed name, but then fails to find the actual name

```
Jun 02, 2021 2:55:45 PM javax.jmdns.impl.DNSIncoming$MessageInputStream readName
WARNING: Bad domain name: possible circular name detected. Bad offset: 0x3fff at 0x230a
Jun 02, 2021 2:55:45 PM javax.jmdns.impl.constants.DNSRecordType typeForIndex
WARNING: Could not find record type for index: 65535
Jun 02, 2021 2:55:45 PM javax.jmdns.impl.DNSIncoming readAnswer
WARNING: Could not find record type. domain: 
```

then the first print, and then

```
un 02, 2021 2:55:45 PM javax.jmdns.impl.constants.DNSRecordClass classForIndex
WARNING: Could not find record class for index: 65535
Jun 02, 2021 2:55:45 PM javax.jmdns.impl.DNSIncoming readAnswer
WARNING: Could not find record class. domain:  type: TYPE_IGNORE index 0
```
and the second print...

note that this log references a machine I do not currently have access to. I cannot do much more debugging.

melloware:
Can this be closed now that the PRs have been merged?

PR #284, opened 2024-08-27
Bring relief on excessive logging from DNSIncoming() class

### LINKED ISSUE #282: Warning level logs from DNSIncoming.readAnswer method
Dear Maintainers of JmDNS library,

Recently we came across a strange issue, and what happened is that time to time huge multiline messages started to appear in our log (attached [jmdns_log.txt](https://github.com/user-attachments/files/16596319/jmdns_log.txt)). 

There were warning level messages from DNSIncoming.readAnswer, [here](https://github.com/jmdns/jmdns/blob/f6d37e369cde11dd23b52341ec3f4f50c704640f/src/main/java/javax/jmdns/impl/DNSIncoming.java#L304) and from [here](https://github.com/jmdns/jmdns/blob/f6d37e369cde11dd23b52341ec3f4f50c704640f/src/main/java/javax/jmdns/impl/DNSIncoming.java#L309).
One could argue whether such huge message is necessary in a warning level log, but thats a [different question](https://github.com/jmdns/jmdns/issues/229).

I honestly don't know the reason why it was written out. **Is the message out of standard?**
The only piece of information what I could grasp here is the type of the service __dosvc._tcp.local._

Probably this is part of the [Delivery Optimization service](https://learn.microsoft.com/en-us/windows/deployment/do/waas-delivery-optimization) of Windows. Maybe I'm wrong, but to me it seems it is a recent development in Windows systems which uses DNS-based service discovery.

Quote from https://learn.microsoft.com/hu-hu/windows/deployment/do/whats-new-do#windows-11-22h2
_"New setting: Local Peer Discovery ..."_

There is another message in the log about "Bad domain name", but the machine - from which the messages are coming - is only part of a workgroup (and not of a domain).

I'm out of ideas. **Do you have any suggestion?**

janossch:
This PR is intended to fix the issue #282 

- _WARNING_ level log changed to  _DEBUG_ level in `DNSRecordClass.classForIndex()`
- _WARNING_ level log changed to  _DEBUG_ level in `DNSRecordType.typeForIndex()`
- Removed `InetAddress` parameter from `DNSIncoming.readAnswer()` to get unified usage of `_packet` field
- `DNSMessage` dump changed from _WARNING_ level to _DEBUG_ level. 
- Call `DNSIncoming.print(boolean dump)` only if debug level logging is enabled (further details: [java:S2629](https://rules.sonarsource.com/java/type/Code%20Smell/RSPEC-2629/)).
- Removed duplicate printouts of DNS entries from `DNSIncoming.print(boolean dump)`. Listing of _questions_, _answers_, _additionals_, etc. moved from `DNSIncoming.toString()` method to the `DNSMessage.print()` method of the base class.
- `DNSMessage.print()` format of the printout has been changed to what was originally in `DNSIncoming` and `DNSOutgoing`

### DIFF
--- a/src/main/java/javax/jmdns/impl/DNSIncoming.java
+++ b/src/main/java/javax/jmdns/impl/DNSIncoming.java
@@ -177,7 +177,6 @@ public final class DNSIncoming extends DNSMessage {
     public DNSIncoming(DatagramPacket packet) throws IOException {
         super(0, 0, packet.getPort() == DNSConstants.MDNS_PORT);
         this._packet = packet;
-        InetAddress source = packet.getAddress();
         this._messageInputStream = new MessageInputStream(packet.getData(), packet.getLength());
         this._receivedTime = System.currentTimeMillis();
         this._senderUDPPayload = DNSConstants.MAX_MSG_TYPICAL;
@@ -217,7 +216,7 @@ public final class DNSIncoming extends DNSMessage {
             // parse answers
             if (numAnswers > 0) {
                 for (int i = 0; i < numAnswers; i++) {
-                    DNSRecord rec = this.readAnswer(source);
+                    DNSRecord rec = this.readAnswer();
                     if (rec != null) {
                         // Add a record, if we were able to create one.
                         _answers.add(rec);
@@ -227,7 +226,7 @@ public final class DNSIncoming extends DNSMessage {
 
             if (numAuthorities > 0) {
                 for (int i = 0; i < numAuthorities; i++) {
-                    DNSRecord rec = this.readAnswer(source);
+                    DNSRecord rec = this.readAnswer();
                     if (rec != null) {
                         // Add a record, if we were able to create one.
                         _authoritativeAnswers.add(rec);
@@ -237,7 +236,7 @@ public final class DNSIncoming extends DNSMessage {
 
             if (numAdditionals > 0) {
                 for (int i = 0; i < numAdditionals; i++) {
-                    DNSRecord rec = this.readAnswer(source);
+                    DNSRecord rec = this.readAnswer();
                     if (rec != null) {
                         // Add a record, if we were able to create one.
                         _additionals.add(rec);
@@ -249,16 +248,19 @@ public final class DNSIncoming extends DNSMessage {
                 throw new IOException("Received a message with the wrong length.");
             }
         } catch (Exception e) {
-            logger.warn("DNSIncoming() dump " + print(true) + "\n exception ", e);
-            // This ugly but some JVM don't implement the cause on IOException
+            logger.warn("Corrupted DNSIncoming message. Enable debug level logging to see the full DNSIncoming() message.", e);
+            if (logger.isDebugEnabled()) {
+                logger.debug("DNSIncoming() dump " + this.print(true) + "\n exception", e);
+            }
+            // This is ugly but some JVM don't implement the cause on IOException
             IOException ioe = new IOException("DNSIncoming corrupted message");
             ioe.initCause(e);
             throw ioe;
         } finally {
             try {
                 _messageInputStream.close();
-            } catch (Exception e) { 
-            	logger.warn("MessageInputStream close error");
+            } catch (Exception e) {
+                logger.warn("MessageInputStream close error. {}", e.getMessage());
             }
         }
     }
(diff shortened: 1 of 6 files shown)


janossch:
Please consider that the PR might fixes the #229 issue as well.

src/main/java/javax/jmdns/impl/DNSIncoming.java
```diff
@@ -289,7 +292,10 @@ private DNSQuestion readQuestion() {
         String domain = _messageInputStream.readName();
         DNSRecordType type = DNSRecordType.typeForIndex(_messageInputStream.readUnsignedShort());
         if (type == DNSRecordType.TYPE_IGNORE) {
-            logger.warn("Could not find record type: {}", this.print(true));
+            logger.warn("Could not find record type.");
```
ViToni, line 295: Please assign `_messageInputStream.readUnsignedShort()` to a variable (maybe `short wireType`) and log this value in the message to help identify the missing / unknown type.

src/main/java/javax/jmdns/impl/DNSIncoming.java
```diff
@@ -301,12 +307,15 @@ private DNSRecord readAnswer(InetAddress source) {
         String domain = _messageInputStream.readName();
         DNSRecordType type = DNSRecordType.typeForIndex(_messageInputStream.readUnsignedShort());
         if (type == DNSRecordType.TYPE_IGNORE) {
-            logger.warn("Could not find record type. domain: {}\n{}", domain, this.print(true));
+            logger.warn("Could not find record type. domain: {}", domain);
```
ViToni, line 310: Please assign `_messageInputStream.readUnsignedShort()` to a variable (maybe `short wireType`) and log this value in the message to help identify the missing / unknown type.

src/main/java/javax/jmdns/impl/DNSIncoming.java
```diff
@@ -301,12 +307,15 @@ private DNSRecord readAnswer(InetAddress source) {
...
+            if (logger.isDebugEnabled()) {
+                logger.debug("DNSIncoming() unknown type: {}\n{}", type, this.print(true));
+            }
         }
         int recordClassIndex = _messageInputStream.readUnsignedShort();
         DNSRecordClass recordClass = (type == DNSRecordType.TYPE_OPT ? DNSRecordClass.CLASS_UNKNOWN : DNSRecordClass.classForIndex(recordClassIndex));
         if ((recordClass == DNSRecordClass.CLASS_UNKNOWN) && (type != DNSRecordType.TYPE_OPT)) {
-            logger.warn("Could not find record class. domain: {} type: {}\n{}", domain, type, this.print(true));
+            logger.warn("Could not find record class. domain: {} type: {}", domain, type);
```
ViToni, line 318: Instead of `type` please log `wireType` (from above comments) as `TYPE_IGNORE` doesn't really help here...

ViToni (changes requested):

src/main/java/javax/jmdns/impl/DNSIncoming.java
janossch: _typeForIndex_() method (before the if statement) which returned the _DNSRecordType_ has already logged (also on _WARN_ level) what you request here. So it would be redundant.

src/main/java/javax/jmdns/impl/DNSIncoming.java
janossch: Could it happen that for a known resource record the record class is unknown? If yes, then it is probably not always `TYPE_IGNORE`, isn't it?

src/main/java/javax/jmdns/impl/DNSIncoming.java
janossch: `DNSRecordType.typeForIndex()` logs this value on WARN level, so it has already been there in the log (similarly as above) . I would like to keep this fix as small as possible.

src/main/java/javax/jmdns/impl/DNSIncoming.java
ViToni: You are right. Just saw it in your logs. (Actually I am not a fan of the logging within `DNSRecordType` but that's a code issue not an issue of this PR.)
This change would still result in two log entries:

```
WARN Could not find record type for index: XYZ
WARN Could not find record type.
```

I'd prefer something more like:
```
WARN Could not find record type. domain: 'abc' type: XYZ
```

This log way also indicates which device might have unknown RR types.

src/main/java/javax/jmdns/impl/DNSIncoming.java
ViToni: Same as above.

src/main/java/javax/jmdns/impl/DNSIncoming.java
ViToni: I see only TYPE_OPT being handled explicitly. So some of the newer RRs might be structured in a way that the record class is used for a different purpose AND would result in `TYPE_IGNORE`.

janossch:
@ViToni I've modified (amend) my last commit (I hope it's not a problem). Please have another look on it!
Maybe (and if you agree) put your old comments to resolved and open new ones if needed, Thank you! 🙏

ViToni (approved):
LGTM.
@janossch Thanks for the time you invested.

kaikreuzer:
@janossch Many thanks for your work on this!
@ViToni Thanks for all your expertise and for reviewing the PR!

PR #285, opened 2024-08-30
Fix issue with visibility of modification in DNSCache and docs cleanup in SimpleLockManager

### LINKED ISSUE #186: OOM and high CPU caused by door camera

ivenhov:
Fix #186 

This is to based on discussions in the comments onwards
https://github.com/jmdns/jmdns/issues/186#issuecomment-2308543051

Related to potential lack of visibility of mutations in the entryLists between threads

### DIFF
--- a/src/main/java/javax/jmdns/impl/DNSCache.java
+++ b/src/main/java/javax/jmdns/impl/DNSCache.java
@@ -217,7 +217,6 @@ public class DNSCache extends ConcurrentHashMap<String, List<DNSEntry>> {
             List<DNSEntry> entryList = this.get(key);
             if (entryList == null) {
                 entryList = new ArrayList<>(3);
-                entryList.add(dnsEntry);
             }
             entryList.add(dnsEntry);
             // re-add to the map to establish happens-before and aid visibility
--- a/src/main/java/javax/jmdns/impl/DNSCache.java
+++ b/src/main/java/javax/jmdns/impl/DNSCache.java
@@ -52,7 +52,7 @@ public class DNSCache extends ConcurrentHashMap<String, List<DNSEntry>> {
 
     private static final long   serialVersionUID    = 3024739453186759259L;
     
-    private final SimpleLockManager _lm = new SimpleLockManager();
+    private final transient SimpleLockManager _lm = new SimpleLockManager();
 
     /**
      *
@@ -218,10 +218,10 @@ public class DNSCache extends ConcurrentHashMap<String, List<DNSEntry>> {
             if (entryList == null) {
                 entryList = new ArrayList<>(3);
                 entryList.add(dnsEntry);
-                this.put(key, entryList);
-            } else {
-                entryList.add(dnsEntry);                
             }
+            entryList.add(dnsEntry);
+            // re-add to the map to establish happens-before and aid visibility
+            this.put(key, entryList);
         }
         return true;
     }
@@ -242,8 +242,12 @@ public class DNSCache extends ConcurrentHashMap<String, List<DNSEntry>> {
                 return false;
             boolean result = entryList.remove(dnsEntry);
             /* Remove from DNS cache when no records remain with this key */
-            if (entryList.isEmpty())
+            if (entryList.isEmpty()) {
                 this.remove(key);
+            } else {
+                // re-add to the map to establish happens-before and aid visibility
+                this.put(key, entryList);
+            }
             return result;
         }
     }
@@ -264,12 +268,12 @@ public class DNSCache extends ConcurrentHashMap<String, List<DNSEntry>> {
             List<DNSEntry> entryList = this.get(key);
             if (entryList == null) {
                 entryList = new ArrayList<>(3);
-                entryList.add(newDNSEntry);
-                this.put(key, entryList);
             } else {
                 entryList.remove(existingDNSEntry);
-                entryList.add(newDNSEntry);
             }
+            entryList.add(newDNSEntry);
+            // re-add to the map to establish happens-before and aid visibility
+            this.put(key, entryList);
             return true;
         }
     }
--- a/src/main/java/javax/jmdns/impl/util/SimpleLockManager.java
+++ b/src/main/java/javax/jmdns/impl/util/SimpleLockManager.java
@@ -1,7 +1,3 @@
-/*
- * Copyright Object Matrix 2017
- */
-
 package javax.jmdns.impl.util;
 
 import java.io.Closeable;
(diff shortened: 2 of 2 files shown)


src/main/java/javax/jmdns/impl/DNSCache.java
```diff
@@ -52,7 +52,7 @@ public class DNSCache extends ConcurrentHashMap<String, List<DNSEntry>> {
 
     private static final long   serialVersionUID    = 3024739453186759259L;
     
-    private final SimpleLockManager _lm = new SimpleLockManager();
+    private final transient SimpleLockManager _lm = new SimpleLockManager();
```
joerg1985, line 55: `transistent` is a keyword to change behavoiur of serialization. This field is final so this is thread safe without any further keyword.

ivenhov: Final fields are also serialized by default, unless static as far as I know.

joerg1985: I did assume this change is for thread safty?

ivenhov: Pull request yes. However I crammed in transient at the same time when I realised DNSCache is Serializable.

src/main/java/javax/jmdns/impl/DNSCache.java
```diff
@@ -218,10 +218,10 @@ public boolean addDNSEntry(DNSEntry dnsEntry) {
             if (entryList == null) {
                 entryList = new ArrayList<>(3);
                 entryList.add(dnsEntry);
```
kaikreuzer, line 220: I assume you also wanted to delete this line?

kaikreuzer (changes requested):

src/main/java/javax/jmdns/impl/DNSCache.java
ivenhov: Sorry Kai, which line you are referring to?

kaikreuzer: The one where I placed the comment: line 220. It looks redundant, since you've added line 222.

ivenhov: Yes, please remove. Good catch, thanks!

melloware:
@ivenhov if you change your first comment text in this thread to `Fix #186` it will link this PR to the ticket

src/main/java/javax/jmdns/impl/DNSCache.java
kaikreuzer, line 220:
```suggestion
```

kaikreuzer: > Yes, corrected.

Where exactly? I still see it...

ivenhov: Number of commits in the pull request changed from 3 back to 2 again. I've accepted yours

melloware:
@ivenhov looks like you need to re-sign the commit

kaikreuzer (approved):
Thanks!

kaikreuzer:
> looks like you need to re-sign the commit

No need, all fine - the DCO check was only red because my change was added and GitHub uses a "co-authored" line on such commits.

PR #292, opened 2024-09-18
Fix superfluous null check in DNSRecord

### LINKED ISSUE #194: Superfluous null handling with StringBuilder.append
For some reason null is being handled manually only to produce the same result as append(null)

https://github.com/jmdns/jmdns/blob/d15acaf8ad5f11fc6a1dfa597e3e36803a332dbd/src/main/java/javax/jmdns/impl/DNSRecord.java#L570

janossch:
Fix #194 
- Remove null check before calling StringBuilder.append() method.
https://github.com/jmdns/jmdns/blob/3499b05b21377c9f7db9de6cd5422d90c69c78c7/src/main/java/javax/jmdns/impl/DNSRecord.java#L570

### DIFF
--- a/src/main/java/javax/jmdns/impl/DNSRecord.java
+++ b/src/main/java/javax/jmdns/impl/DNSRecord.java
@@ -571,7 +571,7 @@ public abstract class DNSRecord extends DNSEntry {
         protected void toString(final StringBuilder sb) {
             super.toString(sb);
             sb.append(" alias: '")
-                .append(_alias != null ? _alias : "null")
+                .append(_alias)
                 .append('\'');
         }
 


melloware:
If you add the line `Fix #194` as the first line of your comment it will link this PR with the ticket.

janossch:
> If you add the line `Fix #194` as the first line of your comment it will link this PR with the ticket.

did you think like this way? ☝️

melloware:
Yep now its linked!

src/main/java/javax/jmdns/impl/DNSRecord.java
```diff
@@ -775,14 +749,14 @@ public boolean isSingleValued() {
         @Override
         boolean handleQuery(JmDNSImpl dns, long expirationTime) {
             ServiceInfoImpl info = (ServiceInfoImpl) dns.getServices().get(this.getKey());
-            if (info != null && (info.isAnnouncing() || info.isAnnounced()) && (_port != info.getPort() || !_server.equalsIgnoreCase(dns.getLocalHost().getName()))) {
+            if (info != null && (info.isAnnouncing() || info.isAnnounced()) && (port != info.getPort() || !server.equalsIgnoreCase(dns.getLocalHost().getName()))) {
```
geroyche, line 752: call be paranoid, but in my book neither the previous version nor the new version is null-safe with regards to server.
same in line 828/802

src/main/java/javax/jmdns/impl/DNSRecord.java
```diff
@@ -683,36 +658,35 @@ protected void toString(final StringBuilder sb) {
...
         public Service(String name, DNSRecordClass recordClass, boolean unique, int ttl, int priority, int weight, int port, String server) {
             super(name, DNSRecordType.TYPE_SRV, recordClass, unique, ttl);
-            this._priority = priority;
-            this._weight = weight;
-            this._port = port;
-            this._server = server;
+            this.priority = priority;
+            this.weight = weight;
+            this.port = port;
+            this.server = server;
```
geroyche, line 677:
```suggestion
            this.server = server != null ? server : "";
```
may be a simply solution

src/main/java/javax/jmdns/impl/DNSRecord.java
janossch: You might be right, but it was never my intention to avoid NPE here. This line is just about renaming fields.

melloware:
@janossch @geroyche is it juts me or locally all the unit tests don't pass?  There is a bunch of PowerMock which is usually a smell for me for unit tests.  A lot of tests pass but i think there are some really ugly ones that fail and will probably always fail.

Do you guys see that locally too?  I run on Windows so not sure if its just me?

janossch:
Yes, I see that too. Currently there are 20 tests which are failing in my environment. Namely:

```
ByteWranglerTest.testTooLongNonAsciiString
ByteWranglerTest.testTooLongString
DNSRecordTest.testDNSRecordKey
JmDNSTest.shouldNotNotifyServiceListenersForServiceResolveAfterServiceRemoved
JmDNSTest.testListenForMyService
JmDNSTest.testListenForMyServiceAndList
JmDNSTest.testListenForServiceOnOtherRegistry
JmDNSTest.testListMyServiceIPV6
JmDNSTest.testQueryMyService
JmDNSTest.testWaitAndQueryForServiceOnOtherRegistry
JmmDNSTest.testListenForMyService
JmmDNSTest.testListenForMyServiceAndList
JmmDNSTest.testQueryMyService
JmmDNSTest.testRegisterService
TextUpdateTest.testCaseSensitiveSubtype
TextUpdateTest.testListenForTextUpdateOnOtherRegistry
TextUpdateTest.testRegisterCaseSensitiveField
TextUpdateTest.testRegisterEmptyTXTField
TextUpdateTest.testRenewExpiringRequests
TextUpdateTest.testSubtype
```
ByteWrangler tests (hopefully) will be fixed by this PR https://github.com/jmdns/jmdns/pull/291

Tests were disabled in CI because https://github.com/jmdns/jmdns/issues/281#issuecomment-2303086485

I hope this will change soon, because it would be great to have some kind of safety net.
I try to follow the strategy that before the PR I execute all tests and I try not to increase the number of failed test.

melloware:
i went back even further to last Decembers release and the tests were still failing then.  I have no idea when the last time the test suite actually passed?

src/main/java/javax/jmdns/impl/DNSRecord.java
janossch: I've applied this suggestion on the commit (amended), I believe this has more benefit, since it could prevent NPE in other methods as well (e.g. in `write()` and `toByteArray()` method of `Service` class)

src/main/java/javax/jmdns/impl/DNSRecord.java
janossch: I've put this to resolved, since below I've applied the other suggestion. Thanks for the hints!

melloware (approved):

src/main/java/javax/jmdns/impl/DNSRecord.java
```diff
@@ -567,31 +555,29 @@ public ServiceEvent getServiceEvent(JmDNSImpl dns) {
...
+                    .append(alias)
+                    .append('\'');
         }
-
     }
 
     public static class Text extends DNSRecord {
-        // private static Logger logger = LoggerFactory.getLogger(Text.class);
-        private final byte[] _text;
+        private final byte[] bytes;
```
geroyche, line 564: i am not a fan of renaming the byte array to bytes instead of text, semantically speaking.
in my view more information was carried when it was called after what it contains, not what it is.

geroyche (approved):
I'm not really in a position to "officially" review this MR, I believe.
having said that, it looks good to me.

src/main/java/javax/jmdns/impl/DNSRecord.java
janossch: I've also struggled how to name it, but I'm open for better suggestions. However generally speaking I share the opinion that the field name should not duplicate the name of its containing class.
https://rules.sonarsource.com/java/type/Code%20Smell/RSPEC-1700/

Of course if this is a blocker of this PR I happy to revert it back to `text`.

geroyche:
So we agree that bytes is not better. In lieu of better options I think you should revert it to text indeed.Am 24.09.2024 um 20:30 schrieb janossch ***@***.***>:﻿
@janossch commented on this pull request.



In src/main/java/javax/jmdns/impl/DNSRecord.java:
>      }
 
     public static class Text extends DNSRecord {
-        // private static Logger logger = LoggerFactory.getLogger(Text.class);
-        private final byte[] _text;
+        private final byte[] bytes;

I've also struggled how to name it, but I'm open for better suggestions. However generally speaking I share the opinion that the field name should not duplicate the name of its containing class.
https://rules.sonarsource.com/java/type/Code%20Smell/RSPEC-1700/
Of course if this is a blocker of this PR I happy to revert it back to text.

janossch:
I didn't say that it isn't better. 😅 IMHO it's better since it has different name than its containing class. Do you have any other suggestion? (Except: `text` and `bytes`)

geroyche:
Text is “just right” because that’s what a dns txt record contains by definition.TXT recorden.wikipedia.orgSo, I don’t have a better suggestion other than “text” or “content”. But I would argue both are better than “bytes” because that’s just the type of the variable.If we decide to store it as a utf-8 String tomorrow, the variable names text or content could remain, while bytes would probably have to be changed.Am 24.09.2024 um 21:50 schrieb janossch ***@***.***>:﻿
I didn't say that it isn't better. 😅 IMHO it's better since it has different name than its containing class. Do you have any other suggestion? (Except: text and bytes)

janossch:
I've changed `bytes` to `content`. Thanks for the suggestion.

geroyche (approved):

melloware:
@janossch looks like there are conflicts now after my latest PR

kaikreuzer:
@janossch I would prefer if you could split this PR into the simple fix itself and other code clean ups in a separate PR. This makes the commit history much cleaner and also helps reviewing the code changes.

janossch:
@kaikreuzer done.

janossch:
@melloware could you please close your PR with the same content? or should i close mine? I will later on submit the other changes in a separate PR.

src/main/java/javax/jmdns/impl/DNSRecord.java
```diff
@@ -571,8 +571,8 @@ public ServiceEvent getServiceEvent(JmDNSImpl dns) {
         protected void toString(final StringBuilder sb) {
             super.toString(sb);
             sb.append(" alias: '")
-                .append(_alias != null ? _alias : "null")
-                .append('\'');
+                    .append(_alias)
```
kaikreuzer, line 574: Why is this changing the indentation?

janossch: It wasn't intentional, but my IDE puts it there.

janossch: Now fixed.

kaikreuzer (approved):
Thanks!

PR #293, opened 2024-09-18
ServiceIInfo defend against NPEs

### LINKED ISSUE #245: IPv6 related NPE in ServiceInfoImpl.getHostAddresses()
jmdns version: 3.5.6. 

I've got a few automated crash reports with this stack trace, with `getHostAddress()` called in a `ServiceListener.serviceResolved()` callback:

```
Caused by: java.lang.NullPointerException
at javax.jmdns.impl.ServiceInfoImpl.getHostAddresses(ServiceInfoImpl.java:420)
at javax.jmdns.impl.ServiceInfoImpl.getHostAddress(ServiceInfoImpl.java:404)
```

looking at `getHostAddresses()`:

```
 @Override
    public String[] getHostAddresses() {
        Inet4Address[] ip4Aaddresses = this.getInet4Addresses();
        Inet6Address[] ip6Aaddresses = this.getInet6Addresses();
        String[] names = new String[ip4Aaddresses.length + ip6Aaddresses.length];
        for (int i = 0; i < ip4Aaddresses.length; i++) {
            names[i] = ip4Aaddresses[i].getHostAddress();
        }
        for (int i = 0; i < ip6Aaddresses.length; i++) {
            names[i + ip4Aaddresses.length] = "[" + ip6Aaddresses[i].getHostAddress() + "]";    
        }
        return names;
    }
```

line 420 is:

```
names[i + ip4Aaddresses.length] = "[" + ip6Aaddresses[i].getHostAddress() + "]";    
```

The only possible thing `null` on that line is `ip6Aaddresses[i]`. 
So I suppose `getInet6Addresses()` can return an `Inet6Address[]` where some elements are `null`.

I have not checked if v3.5.8 would fix this issue but I doubt it. 
Anyway, since my program does not make use of ipv6 at all, I worked-around this issue issue calling `ServiceInfoImpl.getInet4Address().getHostAddress()`.
Still it would be nice to get the bottom line of that NPE.

melloware:
Fix #245: ServiceIInfo defend against NPE's

- [x] Defends against NPE's in the Arrays themselves
- [x] Fixed spelling mistake in the word "Address" it was "Aaddress"

### DIFF
--- a/src/main/java/javax/jmdns/impl/ServiceInfoImpl.java
+++ b/src/main/java/javax/jmdns/impl/ServiceInfoImpl.java
@@ -418,16 +418,40 @@ public class ServiceInfoImpl extends ServiceInfo implements DNSListener, DNSStat
      */
     @Override
     public String[] getHostAddresses() {
-        Inet4Address[] ip4Aaddresses = this.getInet4Addresses();
-        Inet6Address[] ip6Aaddresses = this.getInet6Addresses();
-        String[] names = new String[ip4Aaddresses.length + ip6Aaddresses.length];
-        for (int i = 0; i < ip4Aaddresses.length; i++) {
-            names[i] = ip4Aaddresses[i].getHostAddress();
+        Inet4Address[] ip4Addresses = this.getInet4Addresses();
+        Inet6Address[] ip6Addresses = this.getInet6Addresses();
+
+        // Default to empty arrays if null
+        if (ip4Addresses == null) {
+            ip4Addresses = new Inet4Address[0];
         }
-        for (int i = 0; i < ip6Aaddresses.length; i++) {
-            names[i + ip4Aaddresses.length] = "[" + ip6Aaddresses[i].getHostAddress() + "]";
+        if (ip6Addresses == null) {
+            ip6Addresses = new Inet6Address[0];
         }
-        return names;
+
+        // Use ArrayList to store valid addresses
+        List<String> addresses = new ArrayList<>();
+
+        // Add non-null IPv4 addresses
+        for (Inet4Address ip4 : ip4Addresses) {
+            if (ip4 != null) {
+                addresses.add(ip4.getHostAddress());
+            } else {
+                logger.info("IPv4 address is null");
+            }
+        }
+
+        // Add non-null IPv6 addresses (with brackets)
+        for (Inet6Address ip6 : ip6Addresses) {
+            if (ip6 != null) {
+                addresses.add("[" + ip6.getHostAddress() + "]");
+            } else {
+                logger.info("IPv6 address is null");
+            }
+        }
+
+        // Convert the ArrayList to an array and return
+        return addresses.toArray(new String[0]);
     }
 
     /**


src/main/java/javax/jmdns/impl/ServiceInfoImpl.java
```diff
@@ -418,15 +418,35 @@ public String getHostAddress() {
...
-        for (int i = 0; i < ip6Aaddresses.length; i++) {
-            names[i + ip4Aaddresses.length] = "[" + ip6Aaddresses[i].getHostAddress() + "]";
+
+        String[] names = new String[ip4Addresses.length + ip6Addresses.length];
+
+        for (int i = 0; i < ip4Addresses.length; i++) {
+            if (ip4Addresses[i] != null) {
+                names[i] = ip4Addresses[i].getHostAddress();
+            } else {
+                names[i] = "null"; // Handle null elements as needed
```
kaikreuzer, line 438: The question here is what is "as needed"?
Filling the array with "null" strings looks dangerous to me - a caller of the method probably expects valid IP addresses in the array. I would hope that this else-branch will actually never be entered; but if so, I would rather log a message (maybe INFO level) and not add anything to the array (which will require changes to how the array is constructed and filled).
Wdyt?

melloware: Sure thing i was trying to be as least impactful but I like your idea.

melloware: Switched to INFO messages

src/main/java/javax/jmdns/impl/ServiceInfoImpl.java
```diff
@@ -418,15 +418,35 @@ public String getHostAddress() {
...
+        if (ip4Addresses == null) {
+            ip4Addresses = new Inet4Address[0];
+        }
+        if (ip6Addresses == null) {
+            ip6Addresses = new Inet6Address[0];
         }
-        for (int i = 0; i < ip6Aaddresses.length; i++) {
-            names[i + ip4Aaddresses.length] = "[" + ip6Aaddresses[i].getHostAddress() + "]";
+
+        String[] names = new String[ip4Addresses.length + ip6Addresses.length];
```
kaikreuzer, line 432: This is what I referred to with "which will require changes to how the array is constructed and filled" - your array size is now still the same, but you might have empty entries for certain indexes, which is not good.
Imho, the array should only contain valid IP addresses and have no gaps inbetween, would you agree?

melloware: agreed.  I missed that point.

melloware: OK i took a different approach where it has to count them up front to get the proper size and then add the data into the array

kaikreuzer: Tbh, iterating multiple times over the same lists doesn't look very elegant to me.
How about simply using `names = ArrayList<>()` and converting it to an array before returning the result (`names.toArray()`)?

melloware: Good idea how does it look now?

kaikreuzer: That looks lovely! ❤️

kaikreuzer (approved):

PR #294, opened 2024-09-19
Fix DNSRecordTest

melloware:
Fixed failing DNSRecordTest.

- [x] Switched from PowerMock to Mockito as Mockito can handle mocking the `clone()` method that PowerMock no longer can in later JDK's.
- [x]  Switched visibility of `readAnswer()` from `private` to `package protected` to allow testing.
- [x] Moved the test to the correct package so it can properly access package protected.

cc @janossch

### DIFF
--- a/src/test/java/javax/jmdns/impl/DNSRecordTest.java
+++ b/src/test/java/javax/jmdns/impl/DNSRecordTest.java
@@ -13,26 +13,21 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package javax.jmdns.test;
+package javax.jmdns.impl;
 
 import org.junit.Assert;
+import org.junit.Before;
 import org.junit.Test;
-import org.junit.runner.RunWith;
-import org.powermock.api.easymock.PowerMock;
-import org.powermock.core.classloader.annotations.PrepareForTest;
-import org.powermock.modules.junit4.PowerMockRunner;
-import org.powermock.reflect.Whitebox;
+import org.mockito.Mockito;
 
 import javax.jmdns.ServiceInfo;
-import javax.jmdns.impl.DNSIncoming;
-import javax.jmdns.impl.DNSRecord;
 import javax.jmdns.impl.constants.DNSRecordClass;
 import javax.jmdns.impl.constants.DNSRecordType;
 
-import static org.easymock.EasyMock.expect;
+import static org.mockito.Mockito.*;
 
-@RunWith(PowerMockRunner.class)
-@PrepareForTest({DNSIncoming.class, DNSIncoming.MessageInputStream.class})
+import java.lang.reflect.Field;
+
 public class DNSRecordTest {
 
     private static final int TTL_IN_SECONDS = 60 * 60; // ONE HOUR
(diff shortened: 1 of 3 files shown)


melloware:
cc [joerg1985](https://github.com/joerg1985)

src/test/java/javax/jmdns/impl/DNSRecordTest.java
```diff
@@ -118,12 +117,12 @@ public void testServiceInfoFromDNSRecord() {
     @Test
     public void testDNSRecordKey() {
         DNSRecord record1 = new DNSRecord.Service("android_123.local.", DNSRecordClass.CLASS_IN, true, TTL_IN_SECONDS, 0, 0, 0, "test_server");
-        Assert.assertEquals("_android_123.local.", record1.getKey());
+        Assert.assertEquals("android_123.local.", record1.getKey());
```
janossch, line 120: To be honest, I'm biased, because I like `mockito` ... but this was the ~point~ _underline_ (🥲) when I was unsure.. unsure whether the test is wrong here or the product. :)

melloware: @carlonzo  you wrote this test and it used to have "_android" but I think the change here just broke that: https://github.com/jmdns/jmdns/pull/275 from @joerg1985

melloware: actually I don't see why this ever should have had and "_" to start the name.

carlonzo: Hey thanks for tagging, I'm pretty sure I've seen that coming from an mdns server on Android and was causing issues.

janossch: @melloware 

What I would do then, is to put back the original `Assert` as it was before

```suggestion
        Assert.assertEquals("_android_123.local.", record1.getKey());
```

but modify the line above to create the record with a leading underscore:

```
... DNSRecord.Service("_android_123.local." ...
```
To check whether that _"special"_ leading character is not removed.

melloware: OK test updated for underscore

janossch (approved):
LGTM 👍

melloware:
cc @ViToni this is the dedicated PR to fix the DNSRecordTest

ViToni:
@melloware IMHO it would be preferable to keep changes to the bare minimum.
E.g. changes regarding typos or the version bump of slf4j seem unrelated and should each be at least a dedicated commit if not individual PRs.

melloware:
Fair enough.

While I was in the Pom.xml i was fixing what IntelliJ were reporting as errors. With the only exception being the SLF4J upgrade.

melloware:
Also it's disappointing that the unit test suite doesn't work and from what I can tell hasn't passed in years.

melloware:
Reverted all the unnecessary changes.

kaikreuzer (approved):
lgtm, thanks!

PR #295, opened 2024-09-20
4.0.0: JDK 21 Version with VirtualThreads

melloware:
@kaikreuzer just food for thought for a 4.0.0 branch if you want to create one.  

- [x] Modernize to JDK21
- [x] Fix Maven pom.xml for latest plugins needed for 21
- [x] Use `VirtualThreads` instead of real threads by switching to `Executors.newVirtualThreadPerTaskExecutor()`
- [x] Fix all JDK21 recommendations by IntelliJ


ViToni:
@melloware Does the fix for ˋDNSRecordTestˋ require to move to JDK21?
If it does not, I'd suggest to make to separate PRs.
Maybe moving unrelated changes of the ˋpom.xmlˋ to a dedicated commit/PR would useful as well.

melloware:
The DNS record test has a separate PR for  the main branch already. This was to start a discussion of creating a 4.x branch and yes all the Pom.xml changes are necessary to compile for 21.

ViToni:
@melloware Thanks for clarifying this. Would you mind removing the commit fixing ˋDNSRecordTestˋ from this PR for clarity?

melloware:
Ok let me squash it. I needed it to make sure all tests were passing under JDK21 (minus the ones that have always been failing)

melloware:
Commits Squashed

kaikreuzer:
> just food for thought 

Thanks @melloware, interesting food, so let me give you my thoughts:
- This PR contains a lot of clean-ups and typo fixes that are independent of JDK21 and which you could also provide to the `main` branch as a PR.
- I tend to keep the JDK version usually as low as possible. Being a library that is a dependency for other software, it should not put unnecessary constraints onto such software. Since JDK17 premier support still lasts for two more years, I would continue releasing a JmDNS version for JDK17 until then.
- Maintaining separate versions adds quite some efforts, not only for the release management, but also for the contribution process, since commits need to be cherry-picked to other branches or additional PRs must be implemented for the same fix for the different branches. This is something I'd like to avoid.
- I have no experience with virtual threads of JDK21 so far. Do they bring a huge benefit in terms of lower CPU load or memory consumption? If there are strong reasons for users asking for this, I'd be ok in figuring out a lean way how a JDK21 with virtual threads could be provided additionally to the normal JDK17 version.

melloware:
This explains why VirtualThreads are revolutionary for Java development: https://blog.rockthejvm.com/ultimate-guide-to-java-virtual-threads/#2-why-virtual-threads

I will submit a separate PR on the main branch for the spelling, gramar and JDK8 specific improvements!

Siedlerchr:
I generally like the idea but I am just concerned about the impacts on running this library on Android. Seems like Android does not support virtual threads yet https://issuetracker.google.com/issues/242309109

melloware:
@Siedlerchr i agree this would not help Android that is why i suggested having 3.X (java8) and 4.X (java21) releases.

I use jMDNS in this Quarkus Extension and Quarkus is already fully updated to support Virtual Threads: https://github.com/quarkiverse/quarkus-mdns

ViToni:
We use this library in a larger installation where for technical reasons (processor archtiecture) we are bound to use Java8.
It would be really appreciated to have a Java8 compatible version. AFAIK one benefits already by RUNNING software on more recent JVMs. Newer JVM versions for runtime might be more beneficial than changes which would make code incompatible with older Java version.

In addition I'd like to chime in what @Siedlerchr mentioned:
> I generally like the idea but I am just concerned about the impacts on running this library on Android. 

This is especially true if one has to support also older Android versions. 
Accoridng to this chart https://de.statista.com/statistik/daten/studie/180113/umfrage/anteil-der-verschiedenen-android-versionen-auf-geraeten-mit-android-os/ (the site is German but the numbers should speak for themselves).
Android 11 and below make about on third of active systems.

Java 17 is supported starting with Android 14: https://developer.android.com/build/jdks#compileSdk

My not so old Fairphone 4 has just Android 13, means only support up to  Java 11...

Siedlerchr:
> Java 17 is supported starting with Android 14: https://developer.android.com/build/jdks#compileSdk

Our app uses Kotlin and java and all our libs are compiled with toolchain jdk17, and they run on older versions  (API Level 29) of Android 10 just fine

melloware:
I am going to close this branch and re-open which should have a much smaller changeset for those that want to view it.

PR #298, opened 2024-09-24
JDK8 Cleanup

melloware:
@kaikreuzer as requested this is cleaning up JDK8 warnings and spelling mistakes


kaikreuzer (approved):
Wow, thanks, that's a long list of clean-ups!
I've scanned over it and it all looks good to me - I just hope that we do not accidentially break anything...

PR #301, opened 2024-09-27
Fix #185: ServiceInfo keep property order

### LINKED ISSUE #185: Attribute order not preserved
At present the attribute names returned by `getPropertyNames()` do not reflect the actual order of properties received. So implementations that care about this order cannot see what it was.

This occurs because [ServiceInfoImpl.java creates a Hashtable](https://github.com/jmdns/jmdns/blob/6e69389ee5799861b7d6334020179971ae737ebc/src/main/java/javax/jmdns/impl/ServiceInfoImpl.java#L763) to receive the attribute list. [LinkedHashMap](https://docs.oracle.com/javase/8/docs/api/java/util/LinkedHashMap.html) may be a better choice for this data structure, offering the same efficiency but with ordering guarantees.

melloware:
Fix #185: ServiceInfo keep property order

### DIFF
--- a/src/main/java/javax/jmdns/impl/ServiceInfoImpl.java
+++ b/src/main/java/javax/jmdns/impl/ServiceInfoImpl.java
@@ -15,7 +15,7 @@ import java.util.Collections;
 import java.util.EnumMap;
 import java.util.Enumeration;
 import java.util.HashSet;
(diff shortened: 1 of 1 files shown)


kaikreuzer (approved):
Lgtm, thanks!

PR #303, opened 2024-09-27
Add DNSRecord not null checks

### LINKED ISSUE #106: DNSRecord$Address contructor allows null address but some code assume non-null
```
  protected Address(String name, DNSRecordType type, DNSRecordClass recordClass, boolean unique, int ttl, byte[] rawAddress) {
            super(name, type, recordClass, unique, ttl);
            try {
                this._addr = InetAddress.getByAddress(rawAddress);
            } catch (UnknownHostException exception) {
                logger1.warn("Address() exception ", exception);
            }
        }
```

If there is an UnknownHostException, a log warning with stack trace is displayed and _addr will be null. There is this member functions do not taking into account that _addr (via Address.getAddress()) can be null:

```
 @Override
        protected void toByteArray(DataOutputStream dout) throws IOException {
            super.toByteArray(dout);
            byte[] buffer = this.getAddress().getAddress();
            for (int i = 0; i < buffer.length; i++) {
                dout.writeByte(buffer[i]);
            }
        }
```

It would also be good to audit all the code either accessing _addr directly or via getAddress().

For example ServiceInfoImpl.updateRecord:

```
  @Override
    public void updateRecord(DNSCache dnsCache, long now, DNSEntry rec) {
        if ((rec instanceof DNSRecord) && !rec.isExpired(now)) {
            boolean serviceUpdated = false;
            switch (rec.getRecordType()) {
                case TYPE_A: // IPv4
                    if (rec.getName().equalsIgnoreCase(this.getServer())) {
                        _ipv4Addresses.add((Inet4Address) ((DNSRecord.Address) rec).getAddress());
                        serviceUpdated = true;
                    }
                    break;
                case TYPE_AAAA: // IPv6
                    if (rec.getName().equalsIgnoreCase(this.getServer())) {
                        _ipv6Addresses.add((Inet6Address) ((DNSRecord.Address) rec).getAddress());
                        serviceUpdated = true;
                    }
                    break;
...
```

melloware:
Fix #106: DNSRecord a few places not null checking getAddress()

`getAddress()` can be NULL and a couple of places were not NULL checking

### DIFF
--- a/src/main/java/javax/jmdns/impl/DNSRecord.java
+++ b/src/main/java/javax/jmdns/impl/DNSRecord.java
@@ -213,20 +213,19 @@ public abstract class DNSRecord extends DNSEntry {
 
         @Override
         void write(MessageOutputStream out) {
-            if (_addr != null) {
-                byte[] buffer = _addr.getAddress();
-                // If we have a type A records we should answer with a IPv4 address
-                if (_addr instanceof Inet4Address) {
-                    // All is good
-                } else {
-                    // Get the last four bytes
-                    byte[] tempbuffer = buffer;
-                    buffer = new byte[4];
-                    System.arraycopy(tempbuffer, 12, buffer, 0, 4);
-                }
-                int length = buffer.length;
-                out.writeBytes(buffer, 0, length);
+            if (_addr == null) {
+                return;
             }
+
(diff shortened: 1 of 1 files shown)


kaikreuzer (approved):
lgtm, thanks!

PR #305, opened 2024-09-27
DNSQuestion consistency to only add answers from matching service

### LINKED ISSUE #259: JMDNS responds to all Type A resolve hostname messages
JMDNS will respond to _any_ Type A/AAAA message it sees. For example, if attempting to resolve a particular hostname w/ avahi 

**`avahi-resolve-host-name --name nodeabc.local`**

JMDNS will respond regardless of if it's running on **nodeabc**. 

It seems like there's no check for the Query name:

https://github.com/jmdns/jmdns/blob/020e7ccbcbc3ecd0f0139dd007acb24153e4510e/src/main/java/javax/jmdns/impl/DNSQuestion.java#L37

like there is for service:

https://github.com/jmdns/jmdns/blob/020e7ccbcbc3ecd0f0139dd007acb24153e4510e/src/main/java/javax/jmdns/impl/DNSQuestion.java#L134

melloware:
Fix #259: DNSQuestion consistency to only add answers from matching service

Refactor to prevent duplicate code and check for  host match in both `Address` and `Server` consistently.

### DIFF
--- a/src/main/java/javax/jmdns/impl/DNSQuestion.java
+++ b/src/main/java/javax/jmdns/impl/DNSQuestion.java
@@ -25,6 +25,28 @@ public class DNSQuestion extends DNSEntry {
     private static final Logger logger = LoggerFactory.getLogger(DNSQuestion.class);
 
     /**
+     * Checks if the local host name matches the name of this instance, ignoring case.
+     * If there is a match, adds the DNS records for the local host to the provided set of answers.
+     *
+     * @param jmDNSImpl The {@link JmDNSImpl} instance representing the local DNS service.
+     * @param answers   The set of {@link DNSRecord} to which the local host's DNS records should be added if a match is found.
+     * @return {@code true} if the local host name matches this instance's name (case-insensitive); {@code false} otherwise.
+     */
+    boolean processHostMatch(JmDNSImpl jmDNSImpl, Set<DNSRecord> answers) {
+        HostInfo hostInfo = jmDNSImpl.getLocalHost();
+        String localHostName = hostInfo.getName();
+
+        // Use equalsIgnoreCase directly without lowercasing.
+        if (!localHostName.equalsIgnoreCase(this.getName())) {
+            return false;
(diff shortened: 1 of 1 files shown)


kaikreuzer (changes requested):
I would prefer some refactoring to make the logic flow clearer when reading the code. Could you check if that's possible?

src/main/java/javax/jmdns/impl/DNSQuestion.java
```diff
@@ -24,6 +24,28 @@
...
+        HostInfo hostInfo = jmDNSImpl.getLocalHost();
+        String localHostName = hostInfo.getName();
+
+        // Use equalsIgnoreCase directly without lowercasing.
+        if (!localHostName.equalsIgnoreCase(this.getName())) {
+            return false;
+        }
+
+        // If the names match, add all the answers and return true.
+        answers.addAll(hostInfo.answers(this.getRecordClass(), this.isUnique(), DNSConstants.DNS_TTL));
```
kaikreuzer, line 45: This is imho an unexpected sideeffect of the method. I expect an `isXYZ`method to do the check and return `true` or `false`, but be without any additional effects.

kaikreuzer (changes requested):
I would prefer some refactoring to make the logic flow clearer when reading the code. Could you check if that's possible?

src/main/java/javax/jmdns/impl/DNSQuestion.java
melloware: Sure thing let me try and improve this i was trying to DRY it up with changing as little code as possible.  But let me see if it can be improved.

melloware: You are right i like the code but renamed `isHostCheck` to `processHostMatch` to indicate its more than a boolean that is processing something,

kaikreuzer (approved):
Much nicer, thanks!

PR #312, opened 2024-09-30
Multi-Release JAR with JDK21 Virtual Threads

melloware:
How is this for a compromise.  It uses JDK MultiRelease build to work for JDK8 or add Virtual Threads for 21+.

See: https://www.baeldung.com/maven-multi-release-jars

It was introduced in the JDK: https://openjdk.org/jeps/238


melloware:
cc @ViToni @Siedlerchr

src/main/java/javax/jmdns/impl/util/ExecutorServiceFactory.java
```diff
@@ -0,0 +1,49 @@
...
+public class ExecutorServiceFactory {
+
+    /**
+     * Creates a new single-threaded executor service where all tasks are executed sequentially
+     * in a single thread.
+     *
+     * @param name the name to be used for the thread in this executor
+     * @return a newly created single-threaded {@link ExecutorService}
+     */
+    public ExecutorService newSingleThreadExecutor(String name) {
```
janossch, line 23: Are these required to be instance methods (because of this multi-release jar) or could it be changed to static method like in case of `Executors`? 
▶️ if _yes_ then you don't need to instantiate this factory whenever you need an `ExecutorService`, and it would follow the logic of `Executors`. wdyt?

melloware: I wasn't sure i noticed in a lot of the examples I saw none of the classes being replaced were `static` so i errored on the side of caution but I can try it.  It makes logical sense to be static

melloware: Looks like it works as static!

janossch (approved):
@melloware thx for asking my review opinion. However I cannot judge whether the virtual threads are always (in all circumstances) a good alternative compared to normal threads. In terms of code it looks good to me.

melloware:
Yep so far in my BETA testing i have not seen any anomalies but you know in the real world with all this different devices would be the real test 😄

ViToni:
I know of a large scale OSGi based deployment working on Java8. Id' like to have this change tested there before signing this off.

melloware:
Any chance you could build this SNAPSHOT and test it?

kaikreuzer:
@melloware This is a nice and elegant approach. I assume that the size of the Jar should hardly increase as it is merely a single class that is added in Java21-style, right?
Clearly, this should be tested to work without any hiccups on Java8 systems, so let's wait for @ViToni's test results.

Btw, I see again many "source clean-up changes" within this PR. Maybe you can use the time to extract these into an independent PR, which we could already merge meanwhile and which would make this PR here short and easy to read, review and understand. 😄

melloware:
@kaikreuzer yes we should test thoroughly on a Java8 only system.  

However the source cleanup in the pom.xml was all necessary to compile for V21?   Some of the things I can't fix unless upgrading all those plugins etc to support multimodule?

kaikreuzer:
Well, check out
- https://github.com/jmdns/jmdns/pull/312/files#diff-9c5fb3d1b7e3b0f54bc5c4182965c4fe1f9023d449017cece3005d3f90e8e4d8L9-L50
- https://github.com/jmdns/jmdns/pull/312/files#diff-9c5fb3d1b7e3b0f54bc5c4182965c4fe1f9023d449017cece3005d3f90e8e4d8L222-L223
- https://github.com/jmdns/jmdns/pull/312/files#diff-9c5fb3d1b7e3b0f54bc5c4182965c4fe1f9023d449017cece3005d3f90e8e4d8L232
- https://github.com/jmdns/jmdns/pull/312/files#diff-9c5fb3d1b7e3b0f54bc5c4182965c4fe1f9023d449017cece3005d3f90e8e4d8L276-L286
- https://github.com/jmdns/jmdns/pull/312/files#diff-9c5fb3d1b7e3b0f54bc5c4182965c4fe1f9023d449017cece3005d3f90e8e4d8L182-R211 (apart from the version change)

I'd guess all of this can be already done independent of the multi-release change.

melloware:
OK reverted those changes

Siedlerchr:
I am using this in a gradle project is there a way to "request" the version with jre8 or will it be somehow automatically determined? (It's part of a library that is used in android) that's why I am asking

melloware:
the MANIFEST.MF tells its its multirelease JAR so the JVM when it classloads this JAR will see you are using Java8 and ignore the `java21` code

melloware:
@Siedlerchr report back if you find any different please.

kaikreuzer:
> OK reverted those changes

Thanks. Feel free to create a separate PR for these, so that we can merge them directly and not having to wait for this PR.

melloware:
Did anyone get a chance to test this snapshot in a java8 environment?

janossch:
@melloware I wanted to try out this PR on my setup, what I did is I patched my pull request https://github.com/jmdns/jmdns/pull/318 with the contents of this PR locally. In other words I wanted to execute the fixed tests against this PR.

Since I'm using _IntelliJ_ I had a hard time to configure it, but it seems _IntelliJ_ does not support (out-of-the-box) this multi-release output. See:
https://youtrack.jetbrains.com/issue/IDEA-255308/Maven-import-additional-compile-source-roots-with-multiReleaseOutput-option-as-separate-modules

So whenever I debugged the tests, I always ended up in e.g.: `Executors.newSingleThreadExecutor()` even if I used _Java 21_.
Maybe I've not spent enough time to figure out how it could be set up, but do you have any tricks or tips how can I make this work in _IntelliJ_?

The other thing which I came across - even if it is still used normal thread pools - I needed to add a new test dependency, because some tests are failed (basicall all `JmDNSTest`)
```
<dependency>
    <groupId>net.bytebuddy</groupId>
    <artifactId>byte-buddy</artifactId>
    <version>1.15.4</version>
    <scope>test</scope>
</dependency>
```
And it seems related to this [issue](https://github.com/mockito/mockito/issues/3328)

pom.xml
```diff
@@ -51,19 +51,18 @@ The project was originally started in December 2002 by Arthur van Hoff at Strang
...
 			<groupId>org.slf4j</groupId>
 			<artifactId>slf4j-api</artifactId>
-			<version>2.0.7</version>
+			<version>2.0.16</version>
 		</dependency>
 		<dependency>
 			<groupId>junit</groupId>
 			<artifactId>junit</artifactId>
-			<version>4.13.1</version>
+			<version>4.13.2</version>
```
janossch, line 65: Is this essential for this PR? Because it will introduce a merge conflict for the other PR https://github.com/jmdns/jmdns/pull/318. But ultimately the other PR makes this unnecessary.

janossch (changes requested):

pom.xml
melloware: I can revert that its not necessary

melloware:
You might have to do `mvn clean install` to build the SNAPSHOT JAR and then simply use the SNAPSHOT JAR in a project and if it JDK21 it should pick up the 21 version.  Although as to what IntelliJ does I am not sure.

janossch:
Yeah with a sample project outside JmDNS project it works. Maybe I'm doing something wrong, but if I run `mvn clean install` under Java21 for JmDNS project, I saw that the Java 8 class was loaded for the tests. Why is that? 😮

melloware:
I think its because the classloader has to read the Manifest.MF of the JAR at classloading time to see its a multimodule project.

janossch:
So I'm a little bit concerned how this will be tested if more and more things like this will be added.

melloware:
i think it would have to be an integration test and not a unit test.

janossch:
Could the JDK 21 code be carved out to a different maven module somehow and later package together to a single multi-release module? Similarly to what is here: https://github.com/khmarbaise/mrelease
I'm asking this because it could bring relief on how you unit test it and IDEs could easier understand the code.

melloware:
I know what you mean but most libraries face two real solutions.

1. Separate branch which was my original proposal of a 4.0 that was JDK 21 only but @kaikreuzer did not want to maintain two branches which i totally understand.
2. This multi-module build which was why Java invented it for this scenario where we only need to change a few classes and not maintain two branches.

I don't think there are really any other options.

janossch:
I understand. And don't get me wrong, this solution is way more better than going into different branches and separate jars for different _JDKs_. I'm just saying it would be better to separate the different _JDK_ specific codes into specific maven modules (compile time) and in later (packaging) phase it could be packed together into a single jar.

kaikreuzer:
@ViToni Did you already have a chance to test this in your environment?

melloware:
@kaikreuzer this can wait until after you do the current release if you want. I really want to make sure this doesn't break people!

kaikreuzer:
Oh well, we have so many merged PRs now, there's anyhow gonna be something that breaks. 😉 
But sure, I'll do a new release with what we have. Will change version to 3.6.0 as it's definitely not a patch release.

melloware:
Yep i will have to bump the Mockito version here too since its now blowing up using the legacy Mockito.

janossch:
@melloware starting from mockito 5.3.0, `mockito-inline` was removed, see https://github.com/mockito/mockito/pull/2945

melloware:
I just pushed this with Mockito 5.14.2 seeing how it goes

melloware:
OK 1 single failure

```
DNSStatefulObjectTest.testWaitAndSignal:76 The thread should have finished. ==> expected: <true> but was: <false>
```

melloware:
This looks much better but still failing on Mac: https://github.com/jmdns/jmdns/actions/runs/11518328911/job/32064851954

kaikreuzer:
As openHAB is meanwhile on Java 21 as well, I think this PR becomes now even more interesting to be merged.
@melloware It is still marked as draft - where exactly were we with it?

kaikreuzer:
@melloware One observation: The MANIFEST.MF now lists
```
Bundle-RequiredExecutionEnvironment: JavaSE-21
```
This imho prevents the bundle from being resolved on a Java 8 runtime. Shouldn't we keep it to
```
Bundle-RequiredExecutionEnvironment: JavaSE-1.8
```
or will this cause any issues?

melloware:
You are right it should be 1.8!

kaikreuzer:
No comment from @ViToni for a long time. I'd therefore say that this should be unblocked.
@melloware Are you still around and could help if anything goes south after merging?

I think we should bump the version to 3.7.0 or even 4.0.0, wdyt?

melloware:
Yep this should be 4.0.0 as its a major change

kaikreuzer:
Could you make the version bump directly a part of this PR?

melloware:
Sure I will do it in the morning

PR #317, opened 2024-10-03
Code cleanup in packages under javax.jmdns.impl.tasks.

janossch:
- Remove unnecessary whitespaces
- fix several typos in comments, 
- rename local and field variables (remove leading "_")
- use `@throws` instead of `@exception` in comments
- change the `public` constructor of the `public abstract DNSResolverTask` class to `protected`


melloware (approved):

kaikreuzer (approved):
lgtm, many thanks!

PR #322, opened 2024-10-10
Cleanup and modernize Browser sample

melloware:


melloware:
OK @kaikreuzer i did it in 3 separate commits  Moved then files.  Then updated.

kaikreuzer:
For some reason, the move commit still shows some delete/add operations. Maybe it is simply git that does not handle it correctly, I had such situations as well before already...

melloware:
Yep i believe its because so many lines changed Git enforces it to be a Add and Remove.

FlipWarthog:
I think it's just the GitHub diff display is assuming this PR will be squashed so it's ending up as a add/remove. If it gets merged without squashing it should retain all the history since the moves and edits will still be in different commits.

kaikreuzer:
Sorry, merging the license headers caused now conflicts here. Could you please rebase this PR?

melloware:
OK resolved it and now all 3 are marked "Moved"!

src/sample/java/samples/Browser.java
```diff
@@ -11,14 +11,18 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package com.strangeberry.jmdns.tools;
+package samples;
```
kaikreuzer, line 14: `samples` is imho a weird package name. I had suggested `javax.jmdns.tools` on the other PR, so that we stick to the same package name as for the library code. Does anything speak against that?

melloware: Nope I was just following the current convention of they are all"samples" so I thought the idea was to fix it in a new PR?  Right now half would be samples and have would tools?

melloware:
OK renamed all `samples` to `javax.jmdns.tools`

src/sample/java/samples/Browser.java
kaikreuzer: Ah, I see, thanks! I wasn't even aware of the current `https://github.com/jmdns/jmdns/tree/main/src/sample/java/samples` folder. 😄 
But now the PR at least leave it as it was. We can still discuss what to do with all the files in the samples folder.

kaikreuzer (approved):

PR #318, opened 2024-10-07
TEST ONLY - Improving tests

janossch:
- [x] Upgrade to JUnit5. 
- [x] Remove easymock.
- [x] Add logback-classic for tests.
- [x] Make tests 'green'

I wanted to move the below listed files under `javax.jmdns.impl` package, but it needs to be a separate PR (even if they are moved, git cannot preserve their history).
- _DNSStatefulObjectTest.java_
- _JmmDNSTest.java_
- _ServiceInfoTest.java_
- _TextUpdateTest.java_

**No goal:**
- optimize on tests (performance)
- stabilize tests (I can not guarantee there aren't any `flaky` tests)
- remove duplicate tests
- since more or less all `JmDNS` and `JmmDNS` tests are multi-threaded, it was no goal to remove `Thread.sleep()`

PR #325, opened 2024-10-17
Fix mismatch of Mockito library / JDK version preventing test compilation

### LINKED ISSUE #324: 3.5.13-SNAPSHOT fails to compile test due to mismatch of JDK versions
JmDNS make use of Mockito, specifically Mockito 5.2.0. However Mockito 5.x is supposed to run on JDK 11 or above.
When JmDNS is compiled with JDK 8 usage of Mockito fails due to this mismatch in JDK / class file versions:

```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.3:testCompile (default-testCompile) on project jmdns: Compilation failure
[ERROR] /Users/vtoni/git/jmdns/src/test/java/javax/jmdns/impl/DNSRecordTest.java:[19,19] cannot access org.mockito.Mockito
[ERROR]   bad class file: /Users/vtoni/.m2/repository/org/mockito/mockito-core/5.2.0/mockito-core-5.2.0.jar(org/mockito/Mockito.class)
[ERROR]     class file has wrong version 55.0, should be 52.0
[ERROR]     Please remove or make sure it appears in the correct subdirectory of the classpath.
[ERROR]
```

This issue could be solved by going back to Mockito 4.x (latest ATM is 4.11.0).

ViToni:
Fixes #324.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -52,6 +52,7 @@ The project was originally started in December 2002 by Arthur van Hoff at Strang
 		<project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
 		<maven.build.timestamp.format>yyyy-MM-dd HH:mm</maven.build.timestamp.format>
 		<jvm.version>1.8</jvm.version>
+		<mockito.version>4.11.0</mockito.version>
 	</properties>
 
 	<dependencies>
@@ -75,7 +76,7 @@ The project was originally started in December 2002 by Arthur van Hoff at Strang
 		<dependency>
 			<groupId>org.mockito</groupId>
 			<artifactId>mockito-inline</artifactId>
-			<version>5.2.0</version>
+			<version>${mockito.version}</version>
 			<scope>test</scope>
 		</dependency>
 	</dependencies>


melloware (approved):

janossch:
With due respect, I would not change this. If you check the other PR https://github.com/jmdns/jmdns/pull/318. Tests are completely fine under JDK17 even with a newer mockito version.

melloware:
If JDK8 is what is supported @ViToni is right it should build on JDK8

janossch:
Okay. I thought that the Github Action build some kind of golden standard, which by the way uses '17'. And how you do the build locally thats a different question.

https://github.com/jmdns/jmdns/blob/9175588f5d7bf9db7b2ade60737262512919a3ac/.github/workflows/maven.yml#L27

if the build needs JDK8 than why this is set in the pom?
https://github.com/jmdns/jmdns/blob/9175588f5d7bf9db7b2ade60737262512919a3ac/pom.xml#L54

janossch:
Exactly. The build could use a later _JDK_ (e.g. _JDK17_), but can target a previous version of _JDK_ platform (e.g. _JDK8_). BTW since _JDK9_ there is an option to target older platform versions with the released binary. https://openjdk.org/jeps/247
Probably the above `jvm.version` should be used with the `release` option in the maven plugin instead of `source` and `target`. Ultimately the decider factor here how the release binary gets created. I don't know that, maybe @kaikreuzer could give us some insights here.

Anyhow if you (or even the release workflow) still use _JDK8_ you can put this in the maven command line, this will skip compilation of the tests. `-Dmaven.test.skip=true` [source](https://maven.apache.org/surefire/maven-surefire-plugin/examples/skipping-tests.html)

I just wanted to express that there is no information anywhere in the repository that the build/compilation should work with _JDK8_.
Or the tests should **not** be executed under JDK17. So yes, you are right, there is no 
> hard requirement to use the latest and greatest

but there is no requirement for the opposite as well (at least written down in this repository).

So the one (me or any developer who puts efforts in this project) should rely on what is in the repository i.e. in the Github Actions. And currently the automatic build uses _JDK17_.

ViToni:
JmDNS is used by a variety of projects with a variety of requirements. Some of these projects still rely on Java 8.

Using a newer version of Mockito would block development on JDK8 systems without providing any direct benefit, and I see no good reason to do this intentionally.

melloware:
Agreed let's get this merged @kaikreuzer and then after we switch the loggers to not be final I think we are ready for a new release

kaikreuzer (approved):
Thanks!

PR #326, opened 2024-10-24
Downgrade logback-classic test dependency to be compatible with JDK8

### LINKED ISSUE #324: 3.5.13-SNAPSHOT fails to compile test due to mismatch of JDK versions

janossch:
Fixes #324

@ViToni  @kaikreuzer  @melloware 
With a parallel PR https://github.com/jmdns/jmdns/pull/318 I've introduced a _logback-classic_ dependency which is not compatible with JDK8 builds (similarly as _mockito_)

- Downgrade logback-classic test dependency to be compatible with JDK8
- Upgrade maven-compiler-plugin to 3.13.0 (more info about its [use under JDK8](https://maven.apache.org/plugins/maven-compiler-plugin/examples/set-compiler-release.html#usage-on-jdk-8) and see related [JEP](https://openjdk.org/jeps/247)).

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -51,7 +51,7 @@ The project was originally started in December 2002 by Arthur van Hoff at Strang
 	<properties>
 		<project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
 		<maven.build.timestamp.format>yyyy-MM-dd HH:mm</maven.build.timestamp.format>
-		<jvm.version>1.8</jvm.version>
+		<jvm.version>8</jvm.version>
 		<mockito.version>4.11.0</mockito.version>
 	</properties>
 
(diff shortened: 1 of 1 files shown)


melloware (approved):
LGTM

kaikreuzer (approved):
lgtm

PR #330, opened 2024-10-24
Run tests on windows, mac, linux

melloware:
on Mac OS

```
JmmDNSTest.testRegisterService:64 We should see the service we just registered:  ==> expected: <true> but was: <false>
[INFO] 
Error:  Tests run: 80, Failures: 5, Errors: 0, Skipped: 0
```

### DIFF
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -22,13 +22,14 @@ jobs:
       fail-fast: false
       matrix:
         os: [windows-latest, macos-latest, ubuntu-latest]
+        java: [8, 17]
     runs-on: ${{ matrix.os }}
     steps:
     - uses: actions/checkout@v4
-    - name: Set up JDK 17
+    - name: Set up JDK ${{ matrix.java }}
       uses: actions/setup-java@v4
       with:
-        java-version: '17'
+        java-version: ${{ matrix.java }}
         distribution: 'temurin'
         cache: maven
     - name: Build with Maven
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -17,7 +17,7 @@ on:
 jobs:
   build:
 
-    name: Build on ${{ matrix.os }}
+    name: JDK ${{ matrix.java }} on ${{ matrix.os }}
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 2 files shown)


melloware:
@kaikreuzer looks like Mac they fail on...  are you on a Mac?  cc @janossch

kaikreuzer:
Yes, see https://github.com/jmdns/jmdns/pull/318#issuecomment-2436160001

melloware:
well at least we know why 😄

melloware:
should we disable tests on Mac only for now?

janossch:
Unfortunately not exactly the same results, one is different among the 5 failures:

Mac:
DNSStatefulObjectTest.testWaitAndTimeout

In your result:
JmDNSTest.testTwoMulticastPortsAtOnce

janossch:
@melloware Under JDK8 I've seen similar failures... Maybe that could help to reproduce these failed tests.

melloware:
image

JDK21 is because we need the latest Mockito so all 3 failed.

JDK8 apparently is NOT offered for MacOs on Github Actions

melloware:
looks like the tests consistently fail on Mac but fine on Windows and Linux.

janossch:
Even `mockito-inline` _5.2.0_ is still refer to _1.14.1_ from `byte-buddy` which does not support java 21. 

Only after there release of ASM _9.5_ `byte-buddy` started to support java 21:
https://github.com/raphw/byte-buddy/commit/57e32d779a94f163f2c9f9ee04a276b6588a2cb6

One approach would be to reference this `byte-buddy` version:
```
<dependency>
	<groupId>net.bytebuddy</groupId>
	<artifactId>byte-buddy</artifactId>
	<version>1.14.3</version>
	<scope>test</scope>
</dependency>
```

_Side note:_ `mockito-inline` had been removed, starting from `mockito` _5.3.0_ and in the later versions.
https://github.com/mockito/mockito/releases/tag/v5.3.0

If JmDNS decide to switch to a newer _mockito_ (5+) it would be better to reference `mockito-core` instead of `mockito-inline`.

melloware:
yes we should reference Mockito Core and Add ByteBuddy

kaikreuzer (approved):
Thanks!

Issue #341, opened 2025-06-05
JmDNS fails to update cache for mDNS services that report subtypes correctly. Causes service removal due to TTL timeout.


msmiegal1234:
I have a service that publishes mDNS for itself and also advertises it's service subtype. From wireshark the mDNS messages are similar to this:

_test-mgt._tcp.local: type PTR, class IN, myService-test-image._test-mgt._tcp.local
_abc._sub._test-mgt._tcp.local: type PTR, class IN, myService-test-image._test-mgt._tcp.local
myService-test-image._test-mgt._tcp.local: type SRV, class IN, cache flush, priority 0, weight 0, port 8445, target registry.local
myService-test-image._test-mgt._tcp.local: type TXT, class IN, cache flush
registry.local: type A, class IN, cache flush, addr 192.160.40.20

According to the mDNS spec the two PTR records are not flushable and the SRV, TXT and A records are flushable,

When the JmDNS sends the query to get service status the responses are correct as listed above. The problem is when the JmDNSImpl.handleRecord(DNSRecord record, long now)  method processes the PTR  records it only updates one one of the cached PTR records since the retrieval of the cachedDNS entry does not actually check the subtype when it looks for the match.

The DNSEntry.isSameEntry should also check the subtype so the proper PTR TTL gets updated. If it does not then either the main PTR type is TTL updated or the subtype PTR TTL is updated which will cause the one that is not updated to send a service removed when it should not.

Recommended change:

    /**
     * Check if two entries have exactly the same name, type, subType and class.
     *
     * @param entry
     * @return <code>true</code> if the two entries have are for the same record, <code>false</code> otherwise
     */
    public boolean isSameEntry(DNSEntry entry) {
        boolean sameKey = this.getKey().equals(entry.getKey());
        boolean sameRecordType = this.matchRecordType(entry.getRecordType());
        boolean sameRecordClass = this.matchRecordClass(entry.getRecordClass());
        boolean sameSubtype = this.sameSubtype(entry);

        return sameKey && sameRecordType && sameRecordClass && sameSubtype;
    }

PR #342, opened 2025-06-10
Fix Subtype PTR record TTL updates

msmiegal1234:
JmDNS fails to update cache for mDNS services that report subtypes correctly. Causes service removal due to TTL timeout.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -3,7 +3,7 @@
 	<modelVersion>4.0.0</modelVersion>
 	<groupId>org.jmdns</groupId>
 	<artifactId>jmdns</artifactId>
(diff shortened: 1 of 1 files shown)


janossch:
1.You could squash your 4 commits into a single one, since you touched 1 or 2 files.
2. To fix the DCO check you need to add a simple line to your commit message:
`Signed-off-by: Author Name <authoremail@example.com>`

example:
image

3. if you add the number of the issue in the PR's message (in the first line), it will link them together, e.g.:
`Fix #341`

melloware:
Also please update description field of this ticket to "Fix #xxx" so it links this PR to the ticket

PR #345, opened 2025-06-16
fix checking of unicast or multicast answers

### LINKED ISSUE #343: Request without the QU bit set responds via unicast rather than multicast
I _think_ the error stems from #265.

This change makes the assumption that a unique query means the `QU` bit has been set. I do not believe this is true according to RFC 6762.

I have observed traffic via `tcpdump` of a mDNS request/question without the `QU` bit set being responded to via unicast rather than multicast from a client implementing JmDNS.

hauges:
According to RFC 6762, all Multicast DNS traffic must use UDP port 5353. This applies to the destination port, not the source port. We can be certain that the destination of the DatagramPacket is using port 5353 since we are joining on the group using that port. Therefore, the port check is not necessary.

Fixes #343


melloware (approved):

ViToni:
@hauges I guess you might also need to fix the commit for the check to pass.

kaikreuzer (approved):

PR #346, opened 2025-08-31
Remove dependency on osgi.ee from Manifest

holgerfriedrich:
Bundle-RequiredExecutionEnvironment already defines the minimum Java version. Requiring osgi.ee as capability may cause problems when using this library.

This PR should remove
grafik
from MANIFEST.

@kaikreuzer FYI


kaikreuzer (approved):
lgtm, thanks!

PR #350, opened 2025-08-31
Set minimum RequiredExecutionEnvironment to JavaSE-1.8

holgerfriedrich:
Setting the value to `JavaSE-8` (implicitly introduced in #326) seems to cause issues.
`JavaSE-1.8` works fine for openHAB project.


kaikreuzer (approved):
weird, but so be it...

PR #353, opened 2025-09-22
Fix requestServiceInfo() timeout

Nadahar:
I was looking at the code to try to figure out how to use JmDNS without too much blocking, when I discovered this. As far as I can understand, it has to be a bug.

That said, I don't understand why `requestServiceInfo()` has a timeout at all, since it doesn't return any data and you're supposed to get the results asynchronously via a listener. The "workaround" is to set a really short timeout, which doesn't work for obvious reasons.

### DIFF
--- a/src/main/java/javax/jmdns/impl/JmDNSImpl.java
+++ b/src/main/java/javax/jmdns/impl/JmDNSImpl.java
@@ -891,7 +891,7 @@ public class JmDNSImpl extends JmDNS implements DNSStatefulObject, DNSTaskStarte
      */
     @Override
     public void requestServiceInfo(String type, String name, long timeout) {
-        this.requestServiceInfo(type, name, false, DNSConstants.SERVICE_INFO_TIMEOUT);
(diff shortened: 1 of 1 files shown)


Nadahar:
There are a lot of test failures, but none of them seem to have anything to do with this PR. Some are outdated worker configurations (no longer supported OS/Java combinations), and some seem to me to be timing related (if the test runner runs slightly slower than the test anticipates).

When I try to run the tests locally (on Windows), I get a lot of test failures with `SocketException` - error setting socket option. I see from the Windows GitHub runner that it has the same issues, so I assume that these tests aren't made to work on Windows.

I thus conclude that there's nothing more I can do to address the test failures, so I'll leave this as it is unless I'm asked to change something.

kaikreuzer:
Hey @Nadahar, thanks for the fix!
Never mind the failing tests - I have now removed them with https://github.com/jmdns/jmdns/pull/354 as we never really had them working on macOS (and Java 8 is now deprecated and not available on GHA).

PR #357, opened 2025-12-02
Avoid too many mDNS calls

FFelt:
In the current implementation a call to addServiceListener can cause a recursive call of this method in line 975. This leads to trigger the call to startServiceResolver(type) twice. As a result also the number of mDNS calls in the network are doubled, which leads to unnecessary traffic.

We already use this adjustment to lower the mDNS traffic.


kaikreuzer (approved):
lgtm, many thanks for this fix!

PR #361, opened 2026-04-05
Fix printing JmDNS version

mthmulders:
Before, running the JmDNS JAR would always print

```text
$ java -jar target/jmdns-3.6.4-SNAPSHOT.jar
JmDNS version "RUNNING.IN.IDE.FULL"
```

After this change, it'll print

```text
$ java -jar target/jmdns-3.6.4-SNAPSHOT.jar
JmDNS version "3.6.4-SNAPSHOT"
```


kaikreuzer (approved):
Thanks for the fix!

PR #363, opened 2026-05-17
Fix #356: Windows restore legacy JDK8/11 support

### LINKED ISSUE #356: SocketException thrown when using Java version 1.8 on Windows
# SocketException thrown when using Java version 1.8 on Windows

The following code produces this error when running on openjdk version 1.8.0 on Windows:

## Exception
```
java.net.SocketException: An operation was attempted on something that is not a socket (Error setting socket option)
        at java.net.TwoStacksPlainDatagramSocketImpl.socketNativeSetOption(Native Method)
        at java.net.TwoStacksPlainDatagramSocketImpl.socketSetOption(TwoStacksPlainDatagramSocketImpl.java:146)
        at java.net.AbstractPlainDatagramSocketImpl.setOption(AbstractPlainDatagramSocketImpl.java:309)
        at java.net.MulticastSocket.setNetworkInterface(MulticastSocket.java:554)
        at javax.jmdns.impl.JmDNSImpl.openMulticastSocket(JmDNSImpl.java:493)
        at javax.jmdns.impl.JmDNSImpl.<init>(JmDNSImpl.java:450)
        at javax.jmdns.impl.JmDNSImpl.<init>(JmDNSImpl.java:409)
        at javax.jmdns.JmDNS.create(JmDNS.java:105)
        at xenopus.jmdns.main.JmdnsTestRunner.main(JmdnsTestRunner.java:55)
```

## Code to reproduce
```
public class JmdnsTestRunner {
    private static final String SERVICE_NAME_TCP = "_test._tcp";
    private static final String DOMAIN = "local";
    private static final String TYPE = SERVICE_NAME_TCP + "." + DOMAIN + ".";

    static class SampleListener implements ServiceListener {
        @Override
        public void serviceAdded(final ServiceEvent event) {
            System.out.println("Service added: " + event);
        }
        @Override
        public void serviceRemoved(final ServiceEvent event) {
            System.out.println("Service removed: " + event);
        }
        @Override
        public void serviceResolved(final ServiceEvent event) {
            System.out.println("Service resolved: " + event);
        }
    }

    public static void main(final String[] args) {
        try {
            // add listener
            JmDNS jmdns = JmDNS.create(InetAddress.getLocalHost());  // This is the line throwing the exception
            jmdns.addServiceListener(TYPE, new SampleListener());

            // register service
            ServiceInfo serviceInfo = ServiceInfo.create(TYPE, "example", 1234, "path=index.html");
            jmdns.registerService(serviceInfo);

            System.out.println("Started");

            try {
                Thread.sleep(10000);
            } catch (final InterruptedException e) {
                System.out.println("Interrupted");
            }

            jmdns.unregisterAllServices();
            jmdns.close();
            System.out.println("Done");
            e.printStackTrace();
        }
    }
}
```

## Additional details

This appears to not be an issue in jmdns version 3.6.0, but is an issue in version in version 3.6.1 and 3.6.2. Additionally, using openjdk version 17 resolves the issue.

| jmdns Version | openjdk 1.8         | openjdk 11            | openjdk 17 |
|-----------------|--------------------|---------------------|-------------|
| 3.6.0                | WORKS                | WORKS                 | WORKS      |
| 3.6.1                | Exception thrown | Exception thrown | WORKS      |
| 3.6.2                | Exception thrown | Exception thrown | WORKS      |

### Openjdk version details

The openjdk versions used can be found here: https://learn.microsoft.com/en-us/java/openjdk/download

#### 1.8

```
openjdk version "1.8.0_472"
OpenJDK Runtime Environment (Temurin)(build 1.8.0_472-b08)
OpenJDK 64-Bit Server VM (Temurin)(build 25.472-b08, mixed mode)
```

#### 11

```
openjdk version "11.0.29" 2025-10-21 LTS
OpenJDK Runtime Environment Microsoft-12580167 (build 11.0.29+7-LTS)
OpenJDK 64-Bit Server VM Microsoft-12580167 (build 11.0.29+7-LTS, mixed mode, sharing)
```

#### 17

```
openjdk version "17.0.14" 2025-01-21 LTS
OpenJDK Runtime Environment Microsoft-10800295 (build 17.0.14+7-LTS)
OpenJDK 64-Bit Server VM Microsoft-10800295 (build 17.0.14+7-LTS, mixed mode, sharing)
```

melloware:
Fix #356: Windows restore legacy JDK8/11 support

# Why only Windows?

Because:

```text
TwoStacksPlainDatagramSocketImpl
```

is Windows-specific legacy networking code.

Linux/macOS use different implementations and tolerate this correctly.

---

# Why Java 17 works

Java 17 substantially modernized the socket implementation:

* improved dual-stack handling
* newer UDP implementation
* better multicast behavior
* less reliance on the old TwoStacks native code

So the invalid state no longer occurs.


melloware:
@kaikreuzer this one is safe to merge though

src/main/java/javax/jmdns/impl/JmDNSImpl.java
```diff
@@ -70,6 +70,7 @@
 public class JmDNSImpl extends JmDNS implements DNSStatefulObject, DNSTaskStarter {
 
     private static final boolean IS_WINDOWS;
+	private static final int JAVA_VERSION;
```
kaikreuzer: Use spaces (4) instead of a tab here to match the rest of the file and the project's Eclipse formatter settings (`.settings/org.eclipse.jdt.core.prefs`: `tabulation.char=space`).

src/main/java/javax/jmdns/impl/JmDNSImpl.java
```diff
@@ -369,6 +370,18 @@ public String toString() {
         } else {
             IS_WINDOWS = osName.startsWith("Windows");
         }
+		JAVA_VERSION = getJavaVersion();
+    }
+
+    private static int getJavaVersion() {
+        String version = System.getProperty("java.specification.version");
```
kaikreuzer, line 377: `version` can be null/empty on unusual JVMs — add a guard returning a safe default (e.g. `0`) before calling `startsWith`, consistent with the null-check on `osName` above.

src/main/java/javax/jmdns/impl/JmDNSImpl.java
```diff
@@ -469,7 +482,7 @@ private void start(Collection<? extends ServiceInfo> serviceInfos) {
     }
 
     private InetSocketAddress getMulticastBindAddress(HostInfo hostInfo) {
-        if (IS_WINDOWS) {
+        if (IS_WINDOWS && JAVA_VERSION >= 17) {
```
kaikreuzer, line 488: Add a brief comment explaining why 17: JDK 17 modernized the Windows socket impl, removing the `TwoStacksPlainDatagramSocketImpl` path that throws on a specific-interface multicast bind (see #356).

kaikreuzer (commented):
Reviewing the three minor issues to address. Core fix looks correct and well-scoped — gating the Windows-specific interface bind on JDK 17+ restores legacy JDK 8/11 support without affecting other platforms. Approve once these are addressed.

melloware:
all comments addressed

melloware:
test failure looks unrelated to this change it looks like a timing issue

kaikreuzer (approved):
Many thanks @melloware!
After retriggering the build twice, it finally succeeded. 😉

